PR #7, opened 2018-08-01
changed selectedActive to be inclusive

JunichiOzawa:
I noticed a minor bug when I was looking at the color range demo on heatmap.
https://apexcharts.com/javascript-chart-demos/heatmap-charts/color-range/

when drawing the color range, it is selecting data inclusively but when selecting the legend to filter it, chart excludes the 'to' and 'from' values. this change might have adverse effect on other chart types. I haven't checked it myself yet..

Thanks

### DIFF
--- a/src/modules/Series.js
+++ b/src/modules/Series.js
@@ -75,7 +75,7 @@ export default class Series {
     const selectedActive = function (range) {
       for (let i = 0; i < allHeatMapElements.length; i++) {
         const val = parseInt(allHeatMapElements[i].getAttribute('val'))
-        if (val > range.from && val < range.to) {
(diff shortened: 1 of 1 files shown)


junedchhipa:
If you are getting an error - SVG is undefined, then I think the issue is with the SVG global variable.

The dependency svg.js uses global variable SVG, and it is not working correctly with webpack. I have already contacted the maintainer of svg.js (Fuzzyma), he updated the git repo, but npm still needs to be updated. 

If you want, you can copy the svg.js file from their github repo and get it working temporarily.
More info here:
https://github.com/svgdotjs/svg.js/issues/767

If you are getting some other error, please let me know.
Thanks

PR #134, opened 2018-09-27
Feature/font family

schottey:
Svg.js allows you to set font family of text,
but apexcharts doesn't expose this variable.
This PR allows the user to set their own font family, while keeping Helvetica / Arial as the default.


junedchhipa:
@schottey Thanks for creating PR for font styles. It crossed my mind before but didn't get time to do it.
I will do a review soon.

roquesfr:
Is it possible to add more font-family options?

PR #259, opened 2019-01-09
fix autoScale option

AurelReb:
previous version was not the correct version, sorry for the inconvenience.
_initialize min and max with the first value of selection range
_add null values exception

# New Pull Request

I'm so sorry, I forgot to push my local changes before making the previous pull request.
The previous version was just not working at all (but didn't create new bugs) I tested this version and it works totally!
Sorry again for the inconvenience

Fixes # (246)

## Type of change

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings


junedchhipa:
Ok, no problem. I would like to see a working example (a GIF would work too) to see what this property accomplishes.
Also, there are conflicts with the base branch. Please resolve the conflicts.

AurelReb:
Yes of course. Here is a little gif :
autoscaleyaxis fix

I resolved the merge conflicts

AurelReb:
autoscaleyaxis fix
This one is better

junedchhipa:
Ok, thanks.
I am merging it.
Please create a codepen demo, so I can test with different datasets and multiple series :)

AurelReb:
how can I include the merged version that is not build in codepen/codesandbox ?

junedchhipa:
Thanks, I will check it out.

AurelReb:
just found an issue with syncing charts : the additionnal charts are also changing their yaxis scale like the brush target chart. I thought syncing charts were only sync with xaxis ?

junedchhipa:
Isn't that what this pull request did? You are passing `yaxis` prop also in updateOptions function which is what you wanted to achieve :)

junedchhipa:
If you meant something else, please post a gif or a codepen of the issue

AurelReb:
yes but I want to achieve this only on the target chart of the brush, not
on the other syncing charts (I'm using a combo of brush and sync charts ^^)
So `group` option is syncing xaxis and yaxis options ?

Le jeu. 10 janv. 2019 à 19:42, Juned Chhipa <notifications@github.com> a
écrit :

junedchhipa:
For the moment, you cannot have `grouped` and `brush` charts together.
Well, never really thought about this combination.

One solution is to add a new parameter `updateSyncedCharts` in `_updateOptions` function and while updating targetChart, we can turn that flag off to prevent updating synced charts.

I just committed the idea described above. Please try the fix locally and let me know if it works.

AurelReb:
ok thank you ! the library will be really awesome ;)

AurelReb:
after testing, the syncing chart doesn't update which is... normal with your commit haha
I thought about a workaround to work with sync charts and brush !
Here are my two ideas :

- You can add parametters in `group` option that could say which part of the synced chart should update : 
```javascript
options: {
   chart: {
      group: {
         id : 'name of the group',
         syncedOptions: {
            yaxis: true,
            xaxis: true
         }
      }
   }
}
```
- You can add a multiple target selector in `brush` option:
```javascript
options: {
   chart: {
      brush: {
         enabled : true,
         targets: ['chart#1', 'chart#2']
      }
   }
}
```

I think first idea is more versatile but the second idea is easier to implement.
If I have time this weekend, I can help implement one of them

junedchhipa:
Ah! silly me 😁 
I would go with the second idea, as there can be cases where we might target more than 1 chart.

AurelReb:
Hi ! I just pushed a branch in my fork with the beginning of multiple targets option.
https://github.com/AurelReb/apexcharts.js/commit/9454ba92022f5a8bf2c864fa14903de9b98e2a58
There is something I don't understand : only the last target referenced in the array is updated whereas I make a forEach of the array. Any idea ?

You can test it with the code in [this codesandbox](https://codesandbox.io/s/nrk6ro3krm)
(you just need to test it locally with my changes and by replacing the target in brushOptions by the commented targets)

junedchhipa:
I will take a look.
Btw, I have reverted my commit and removed that extra argument which I provided in _updateOptions

AurelReb:
Un my case, the extra argument you provided in _updateOptions can be useful
if there is a condition in brushSetup so that if there is multiple targets,
the grouped charts are not updated with the brush of one.
So multiple targets allow us to use autoScaleYaxis in different charts and
the "group" option allow to sync cursor selection in these charts.

junedchhipa:
Ah! You might put it back then ;)

AurelReb:
No problem ;) that's why I didn't merge your last commits in my test branch

AurelReb:
> There is something I don't understand : only the last target referenced in the array is updated whereas I make a forEach of the array. Any idea ?

Hi, do you catch this issue ?
I will take a look next Week but really don't understand

junedchhipa:
Couldn't get a chance to look into it, as was busy scheduling the major update 3.0 :)
I will look soon this weekend

AurelReb:
Hi, I suspect the issue to be due to _updateOptions which is a promise. Only the last target is updated, by overwriting the other targets updates

junedchhipa:
Hey, sorry again. Couldn't look into the issue lately as I have been busy with other things.
I will look into `_updateOptions` method to see if it is the root cause. Thanks for the cue.

junedchhipa:
I looked into the issue and I have decided to rewrite the autoScale functionality as it is causing issues in multiple axes. Will update this thread when it's done.

AurelReb:
thanks :smiley:
I'm looking forward to seeing that :tada:

PR #266, opened 2019-01-11
Re-calculate tooltip dimensions on text draw

### LINKED ISSUE #264: Tooltips are going out of bounds when initially triggered on mobile (sometimes on desktop as well)
## Codepen

No Codepen necessary since the problem is reproducible on a demo page as well:
https://apexcharts.com/javascript-chart-demos/line-charts/basic/ 

## Explanation

On mobile (but it happened on desktop as well), when I initially trigger a tooltip for a rightmost (or close to rightmost) data point, it should not go out of view
- What is happening instead?
The tooltip is going out of view on initial trigger, and only after triggering another tooltip and coming back to the original one is the issue fixed
None

screenshot_20190110-112111_chrome

OriginalEXE:
# New Pull Request

As per #264 , on initial tooltip positioning it goes out of bounds. This happens because tooltip dimensions are calculated and then cached before the text in it is updated. This pull request makes sure the dimensions are re-calculated in case the text is updated. Ideally, this would happen in a more structured way (this way the calculation happens twice), but I believe that would require some major re-structuring of the code.

Fixes #264 

## Type of change

## Checklist:

### DIFF
--- a/src/modules/tooltip/Labels.js
+++ b/src/modules/tooltip/Labels.js
@@ -41,6 +41,12 @@ class Labels {
       ttItems,
       shared
     })
+
+    // Re-calculate tooltip dimensions now that we have drawn the text
+    const tooltipEl = this.ttCtx.getElTooltip()
+
+    this.ttCtx.tooltipRect.ttWidth = tooltipEl.getBoundingClientRect().width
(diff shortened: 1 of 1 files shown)


junedchhipa:
Thanks @OriginalEXE 
Merged the PR

Issue #262, opened 2019-01-09
Issue with code quality and build


cstlaurent:
Hi,
I was working on investigating the issue #260 with @mamasselin but a couple of issues quickly blocked me after cloning the repo and looking at the code.

Here is a list of issues I found:

### Code quality issues

-    Some tests do not pass on the latest master branch, some e2e tests look flakey and break sometimes, 2 unit tests fail consistently.
-    There is an eslint file, but it is not enforced. It looks like there are more than a thousand issues regarding that.
-    There is a couple of hardcoded/copy-pasted dependencies in the src folder.

### Build problems

-    There is a mixed usage of require and import which causes issues with Webpack.
-    There are Gulp, Webpack, and Rollup used for builds, probably causes a lot of unnecessary maintenance burden (was there a reason for that?)
-    I think the ESM/module build is misconfigured. If ApexChart is imported through NPM, the Webpack bundler always take the dist/apexcharts.min.js, making it very hard to debug in dev mode and optimize correctly for production mode.
-    Since some dependencies are hardcoded in the project (svg.js, some polyfills) and not correctly imported through NPM, Webpack is not able to optimize them and deduplicate/tree-shake dependencies.
-    Dev dependencies like rollup and Webpack are quite old and should probably be updated if kept at all.
-    Tests and Lint are not run in the CI, making it very hard to do proper gate-keeping in PRs.
-    Changelog or release definitions are not properly kept up to date, making it hard for an external contributor and users to see what was changed or fixed.

This may seem like much negativity, but even with those issues, I still think your library is one of the best out there, and I thank you for creating it.

I think my colleague @mamasselin and I can help out on most of those issues if you are open to tackling them.

I can open separate issues for every point and elaborate a plan to fix them or propose fixes if you are ok with that.

Thanks!

junedchhipa:
@cstlaurent 
Those are very good points, and I agree on many of them. I will try to answer each of them.

1. Mixed usage of import and require is due to the SVG.js not allowing import in their previous versions. They released v3.0 recently, and I haven't upgraded to it.
2. Why Webpack, gulp, and rollup?
Initially, Webpack was used primarily for the build. Later on, a feature request to expose ApexCharts as ES module was made. Webpack didn't help much in those regards (maybe because I am using older versions), so I used rollup.
Gulp, on the other hand, does some minor tasks of copying files and cleaning some logs. If you have alternatives for it, I am open to implementing.
3. Hardcoded dependencies.
[...]

PR #270, opened 2019-01-11
WIP: Build improvements

cstlaurent:
Hi,

This is a follow-up PR for the issue #262 addressing some of the mentioned topics.

### Improvements

- For better code standardization, Eslint was properly configured with Prettier which allows automatic formatting of code in the future. (I did not apply the Eslint fixes in this PR since it would be hard to understand what has been done. Will do in future PR.)
- Build pipeline has been reduced from Webpack, Gulp, and Rollup to only Rollup. The new pipeline was mostly inspired by what is done in the vue.js project.
- The build now creates proper UMD, UMD + minified, ESM and CJS output files with proper package.json configurations. This modification allows all usage scenarios, UMD can be used directly in browsers, ESM with the latest build tools and CJS with older ones.
- Package.json entries were added for jsdelivr and unpkg. This allows CDNs to choose the proper UMD build output.
- Dev dependencies were cleaned up and locked to a specified version (current latest) so that it is possible to have reproducible builds.
- All `require` and `module.exports` have been removed and converted to ecmascript modules `import/export` in all src and tests folders. The mix and match use of cjs and esm was probably the root cause of all build issues.

### Issues/Questions

- There seems to be a limitation/bug with Babel 7 which causes some tests, and the edge/IE build to fail. I’m still investigating the source of the issue.
- I’m not sure how the locales files are dealt with… I don’t see any code/way to load them dynamically, so I’m not sure what to do with them with the build.
- Some npm scripts have changed, but you would still use npm run build to generate the output.
- I’m not sure how you create new releases, so I did not commit the new files under the dist/ folder, so they would need to be generated for the new version.

There seem to be no breaking changes, but I think creating a pre-release npm package would be better so we can test all the packages components correctly using npm/yarn and validate that everything works as it should.


junedchhipa:
Excellent work!
I just reviewed the commits and going to merge it. I will also look into the IE build issue.
Thanks.

junedchhipa:
For new releases, I follow the simple steps.
```bash
npm run build
git commit - new build > push
npm version patch|minor|major
npm publish
```

junedchhipa:
After trying fixes for IE11 for several hours, injecting polyfills and finally getting it to work, I decided to abandon IE11 support and put a notice in the README to include a polyfill if IE11 support is required.
As to including polyfills increased the minified size by 90 KB, I think is not worth it just for IE11.

Hence, the next release will be a major update 3.0.
Everything else worked really well

cstlaurent:
Thanks for merging the PR!

I think its the way to go too for IE. Having a 90 KB size/parse hit for everyone just to support it is hard to justify.

junedchhipa:
@cstlaurent I think, the new build configuration has led to an error on Windows Environment as seen on this issue #278 
Can you please test `npm install` on Windows?

cstlaurent:
Yeah, it looks like the install is building the native module node-sass for node v11 because there is no native build packages for < 4.9 as seen here: https://github.com/sass/node-sass

I will look into what is referring this package and causing the issue.

cstlaurent:
Well from what I can see there is no dependencies on node-sass, this is probably caused by another package.

image

Issue #263, opened 2019-01-09
Feature Request: Range annotations / Highlight region


mamasselin:
It would be cool to have the ability to create "range annotations" to display regions on the chart. It would use min/max values instead of a single value. 

Maybe also a fillColor/fillOpacity to add flexibility, or simply re-use the current borderColor with lowered opacity as the fill color.

Example:

![image](https://user-images.githubusercontent.com/16145977/50917898-c3e15b80-140c-11e9-8507-6946f197c3df.png) 

Ref: https://github.com/chartjs/chartjs-plugin-annotation

junedchhipa:
Good suggestion. This goes to the roadmap.

lixaotec:
Dear fellows,

Was about to open a issue to question this.

Would be nice to have this on blocks, as in Dygraphs library.

![image](https://user-images.githubusercontent.com/40803484/51678928-b0192600-1fc4-11e9-8807-03f9cb960ad8.png)

Dygraphs is a great tool for time series, but lack of vue support, I wish to find all their power in here. Any toughts?

Thanks

junedchhipa:
I am also a fan of Dygraphs and I take inspiration for some of the unique features it provides when it comes to interaction.

There will be some difference in terms of performance and look as Dygraphs is canvas based and ApexCharts is SVG based, but I'll try to integrate the good parts of it into ApexCharts if they seem feasible to implement.

Thanks for the suggestion.

mamasselin:
Hi! I created a PR for this feature request. Please check it out and let me know if anything!

junedchhipa:
Closing this. I will soon add examples in the website as well as write docs for it. Thanks again @mamasselin

PR #338, opened 2019-02-20
Added support for range annotations

mamasselin:
Added support for range annotations feature request (#263):
- New properties for annotations:
    - `x2`: second value used to draw a region instead of a simple line for x-axis annotations
    - `y2`: second value used to draw a region instead of a simple line for y-axis annotations
    - `opacity`: opacity of the region (default 0.3)
- Existing `borderColor` propery is used for the region's fill color with the provided (or default) opacity
- User can enter `x`, `x2`, `y` and `y2` values in any order (e.g.: `x: 200, x2: 400` will produce the same result as `x: 400, x2: 200`)

## Screenshots taken from the sample files included in this PR:

image

## Type of change

## Checklist:

    - [x] No tests but added 2 new sample files

## Addional comments:

- 1 Existing unit test does not pass locally but I'm pretty sure it's not because of my changes. Looks like it has something to do with local time conversion:

    image

- Not sure if normal but the `dist` folder is checked in.


junedchhipa:
Thanks @mamasselin 
Nice work, I will merge soon as everything looks good.

The `dist` folder is checked in and not in .gitignore, but it should be discarded while committing.
Also, you're right, the test doesn't take into account the timezones, will fix it.

 Thanks again for the contribution.

Issue #260, opened 2019-01-09
Bar chart rendering is causing a lot of "Style recalculation"


mamasselin:
## Description

Bar chart rendering is causing a lot of "Style recalculation". The style recalculation seems mostly caused by the `getBBox` call.

For a simple layout:

- 730 datapoints
- Screen is locked for about 1.5 seconds

![image](https://user-images.githubusercontent.com/16145977/50910366-a952b680-13fb-11e9-9f68-3496d379669d.png)

Zoomed to one "Style recalculation" span:

![prolier-details-alone](https://user-images.githubusercontent.com/16145977/50909205-5d067700-13f9-11e9-8950-069a2258a7da.png)

It is even worst when having a chart inside a Vuetify `v-card` (I assume because of the more complex layout).

- 730 datapoints
- Screen is locked for about 10 seconds

![image](https://user-images.githubusercontent.com/16145977/50910588-1bc39680-13fc-11e9-846d-dc916ebecfd8.png)

Zoomed to one "Style recalculation" span:

![profiler-details-in-card](https://user-images.githubusercontent.com/16145977/50909195-5a0b8680-13f9-11e9-8961-54d91662b0dc.png)

## Artifacts

Example: https://codesandbox.io/s/ppvxyl6p57

Profiling results: [apex-profiler.zip](https://github.com/apexcharts/apexcharts.js/files/2741468/apex-profiler.zip)

junedchhipa:
`getBBox` is a function of SVG.js. 
If it is causing performance issues, I will try some native alternatives.

mamasselin:
Please check out PR #347

junedchhipa:
Closing this for now. 
I am going to update to the latest version of `svg.js` where they have performance improvements.

PR #347, opened 2019-02-22
Added check to prevent dataLabels calculation if the "dataLabel" option is disabled

mamasselin:
After further investigation of issue #260 , I have identified the bottleneck causing performance issue for bar charts.

In the `calculateDataLabelsPos` method of `charts\Bar.js`, there is a call to the `getTextRects` method to calculate the size of the dataLabels. This is what makes the performance to drop when having a lot of bars to display.

As a quick workaround, I added a check for whether the `dataLabels` option is enabled in the first place. If not, we simply skip this method altogether, preventing performance drop when not needed.

## Type of change

## Checklist:

## Additional Comments

In addition to this, I was thinking of adding a threshold (arbitrary value) defining a maximum number of bars to allow DataLabels. If the maximum is reached, it could do one of the following:

- log a console warning in the console (e.g.: "DataLabels are enabled but there are too many to display. This may cause performance issue when rendering")

- disable the datalables automatically, with a console warning (e.g.: "DataLabels are enabled but there are too many to display. To avoid performance issue, dataLabels have been disabled for this chart.")

Note that I did not include this in this PR, it's just a suggestion to discuss about.

### DIFF
--- a/src/charts/Bar.js
+++ b/src/charts/Bar.js
@@ -10,6 +10,10 @@ import DataLabels from '../modules/DataLabels'
  * @module Bar
  **/
 
+const DATA_LABELS_WARNING_THRESHOLD = 50
+const DATA_LABELS_WARNING_TEXT =
+  'WARNING: DataLabels are enabled but there are too many to display. This may cause performance issue when rendering'
+
 class Bar {
   constructor(ctx, xyRatios) {
     this.ctx = ctx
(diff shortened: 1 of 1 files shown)


junedchhipa:
Thanks for debugging the issue and creating PR. 
Regarding the warning, the first option seems reasonable - displaying a console.warn when there are too many data-labels.

mamasselin:
ok I will add that to this PR

mamasselin:
I added the console.warn call with threshold and warning text in const at the top of the file.

50 still has good performance. We could even go more than that but the labels don't really display well anyways as the bars are so small when you have that many.

Let me know if anything.

Thanks!

junedchhipa:
Awesome, looks fine!

Issue #352, opened 2019-02-26
Stacked charts in timeseries with missing values are incorrectly displayed


mamasselin:
Stacked charts are not stacking properly when having timeseries with missing values. This was partially addressed as per #199 and #225.

However, issue is still present with latest version.

## Codepen

https://codepen.io/mamasselin/pen/vPEdYR

## Explanation

- What is the behavior you expect?

    I would expect the bars to stack correctly

- What is happening instead?

    Bars are overlapping each others and don't seem to stack properly. Not that this also happens with stacked area charts.

- What error message are you getting?

    None.

mamasselin:
I've started playing around with the stacked line/area and made a fix similar to what was in #353 . However the hover of the points seem broken, but I'm not sure why though.. Is there a separate logic for the points/hovering ? See below

![2019-02-27_09-12-28](https://user-images.githubusercontent.com/16145977/53496496-2ac7dc00-3a70-11e9-9e13-263af35d662e.gif)

mamasselin:
Ok, I think it's just because the datapoints were too close to each other. If I bump the value of 8, the hover works as expected:

![2019-02-27_09-20-11](https://user-images.githubusercontent.com/16145977/53496900-fc96cc00-3a70-11e9-8972-44476abe2189.gif)

junedchhipa:
The function to determine the closest values is named as "getNearestValues" and is located in `src/modules/tooltip/Utils`.
I will take another look, the function became very complex and needs some refactoring.

junedchhipa:
Closing this until I decide to start working on it again.

AlexandraPechueva:
Hi!  We are using in our project ng-apexcharts 1.1.0 (apexcharts 3.10.0 ) and there is a problem with missing timeseries in bar stacked charts, which is fixed in  #353 pull request. When could you include those fixes into release?

iProGamer:
bump

PR #353, opened 2019-02-26
Reviewed logic of calculating stacked bars in timeseries to account for inconsistent X values

mamasselin:
This is a potential fix for the stacked charts issue (#352).

The idea is to check for all series for a corresponding X key instead of going only by index and previous index as it is now.

This way, even when having inconsistent X values, it will still find matching ones and stack accordingly.

### Before:

image

### After:

image

## Type of change

## Checklist:

    - [x] No new tests but I've included a new sample file in this PR

## Additionnal comments

- This particular PR only handles stacked **bar/column** currently. I still decided to submit it to get quick feedback and review on this. If accepted, we can then proceed to tackle other stacked chart types. Let me know.

### DIFF
--- a/samples/vanilla-js/column/stacked-column-empty-points.html
+++ b/samples/vanilla-js/column/stacked-column-empty-points.html
@@ -1,0 +1,91 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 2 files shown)


junedchhipa:
Thanks for the PR @mamasselin 
I will review soon.

PR #356, opened 2019-02-27
Adjusted logic for calculating stacked area/line charts with timeseries, to account for inconsistent X values

### LINKED ISSUE #352: Stacked charts in timeseries with missing values are incorrectly displayed

mamasselin:
Fix similar to what's in #353 but this time for stacked line/area charts

The idea is to check for all series for a corresponding X key instead of going only by index and previous index as it is now.

This way, even when having inconsistent X values, it will still find matching ones and stack accordingly.

Fixes #352

### Before:

image

### After:

image

## Type of change

## Checklist:

    - [x] No new tests but I've included a new sample file in this PR


mamasselin:
Will play around a little more before making this PR "Ready for review"

mamasselin:
After further investigation I don't think this is an acceptable fix. There are other places that break when having inconsistent X values, and fixing it just at this place doesn't really make sense in my opinion. I think if such a fix has to be done it should be made at the "core" level and not repeated for every chart types and all possible scenarios. Otherwise it will get messy and error-prone.

In my project I ended up "normalizing" my data before sending it to apexchart to avoid inconsistent X values, and it works just fine.

Example:

Original data:

```javascript
[ ['2001-01', 15], ['2001-02', 22], ['2001-04', 55] ]
[ ['2001-02', 55], ['2001-03', 17], ['2001-05', 36] ]
```

Normalized data:

```javascript
[ ['2001-01', 15], ['2001-02', 22], ['2001-03', 0], ['2001-04', 55], ['2001-05', 0] ]
[ ['2001-01', 0], ['2001-02', 55], ['2001-03', 17], ['2001-04', 0], ['2001-05', 36] ]
```

Therefore I would rather abandon this PR until a proper solution is found.

Similarly, I would suggest to revert #353 as it does not "completely" fix the issue and may cause unexpected behaviors.

junedchhipa:
@mamasselin Thanks for the update on this.
Yeah, it's a tricky one - I have tried tackling this before but it required a good amount of rework for stacked logic, hence I also abandoned it after a couple of tries.

Thanks for the effort, will re-visit the code when I get enough time to work on the `stacked` logic.
Will revert #353 as suggested by you.

PR #359, opened 2019-02-28
Make `fontFamily` and `type` optional

mikaelkaron:
# New Pull Request

Looking at the [type definition](https://github.com/apexcharts/apexcharts.js/blob/master/types/apexcharts.d.ts#L62) I can see that `fontFamily` is required, but [here](https://github.com/apexcharts/apexcharts.js/blob/master/src/modules/settings/Options.js#L234) it looks like this should be an optional property as it has a default.

The solution

Marked property optional in `apexcharts.d.ts`

## Type of change

## Checklist:

### DIFF
--- a/types/apexcharts.d.ts
+++ b/types/apexcharts.d.ts
@@ -56,7 +56,7 @@ declare module ApexCharts {
 type ApexChart = {
   width?: string | number;
   height?: string | number;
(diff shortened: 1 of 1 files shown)


mikaelkaron:
Realized the same thing goes for `type`, should probably _also_ be optional

PR #415, opened 2019-03-20
Failed to execute 'removeChild' on 'Node'

rantecki:
I'm trying to use ApexCharts inside a Preact app (using react-apexcharts and preact-compat).  Charts render fine, but I get the following error whenever chart.destroy() is called (via componentWillUnmount()):

```
apexcharts.js:26543 Uncaught (in promise) DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at window.removeResizeListener (http://localhost:3333/vendor.js:27035:49)
    at ApexCharts.destroy (http://localhost:3333/vendor.js:27756:16)
    ...
```

It appears that the resize triggers are already gone by the time removeResizeListener is called.  I'm not 100% sure if this is related to Preact, but I'm assuming that if nobody else is having this problem then it must be.  To work around this and related errors I've added several guards around the element modifications in DetectElementResize.js.  It now runs without error.  I don't think these additional checks should cause any interference with regular non-Preact usage.

If you don't wish to support Preact, then fair enough, I guess I can just maintain a private fork.

FYI - I tried to make a codepen for this issue, but I couldn't work out how to use react-apexcharts within codepen (any examples?).

## Type of change

## Checklist:


junedchhipa:
Thanks for creating the PR.
This part is very tricky and needs thorough testing to ensure it doesn't break anything.
I will review soon.

fmflame:
What is going on with this? I am seeing the same issue and this was so much time ago...

junedchhipa:
@fmflame Can you reproduce the issue?
Please do and I will take another look at what's going on.

fmflame:
Yes I was able to reproduce it at the time, that's why I wrote here.

junedchhipa:
Please post a working codepen/codesandbox demo so I can reproduce it on my side.

cyclops26:
Thanks @fmflame I made the same changes to a local fork and it resolved the exact same issue for me in my Preact project including rebasing off current apexcharts.js master.

@junedchhipa I will try to find some time to throw together a test project for you.

PR #687, opened 2019-06-18
fix multiple brush targets

AurelReb:
# New Pull Request

This commit fixes 'multiple brush targets' feature. Before it updated only the last target in the array of targets because the brush selection function was defined to update the last target in the array. Now this function updates each target of the array.

Fixes AurelReb@9454ba9 committed by @junedchhipa in january

## Type of change

## Checklist:


junedchhipa:
Thanks for the PR.

Can you please re-add the images which you deleted? Those images are used in dashboard samples.

AurelReb:
oh sorry, it was a not voluntary deletion.
Btw, I added back your commit which fixed the sync with grouped charts.

junedchhipa:
Thanks!
Looks good.

AurelReb:
Just noticed that when we add forceNiceScale on the brush chart, the autoscale on target charts is also with nice values! The result is perfect, Thank you

Issue #840, opened 2019-08-12
reset zoom at sync line charts


HesamZamanpour:
The problem occur while you follow described scenario at bellow:
1- create some sync line charts or use from this link : https://apexcharts.com/javascript-chart-demos/line-charts/syncing-charts/
2- zoom in one of sync line charts
3- zoom in another one of sync line charts
4- click on reset zoom button of one of charts

the problem : 
the chart that zoomed at third step of scenario do not back to its first xaxis range

recommended solution:
I think changing "revertDefaultAxisMinMax" function to bellow will fix this issue
"
var t = this.w;
var min = this.opts.xaxis.min || (Apex.xaxis && Apex.xaxis.min)
var max = this.opts.xaxis.max || (Apex.xaxis && Apex.xaxis.max)
t.config.xaxis.min = min, t.config.xaxis.max = max, t.config.yaxis.map(function (e, i) { t.globals.zoomed && void 0 !== t.globals.lastYAxis[i] && (e.min = t.globals.lastYAxis[i].min, e.max = t.globals.lastYAxis[i].max) })
"

yingchauw:
React-wrapper ( react-apexcharts 1.3.3 ) seems to have similar issues. 
When I update the options ( the value of x-axis min/max from undefined to timestamp value )  for the timeline chart after componentDidMount(). And then I check the reset button, I saw it triggered zoomed and found min/max were also undefined. Might I seek from your help? Thanks HesamZamanpour

yingchauw:
React-wrapper ( react-apexcharts 1.3.3 ) seems to have similar issues.
When I update the options for the timeline chart after componentDidMount().

And then I click the reset button without other's actions, the chart triggered zoomed functions and the value of min/max in x-axis were also undefined. Might I seek from your help? Thanks.

CodeSandBox
https://codesandbox.io/s/react-basic-example-4snwv

PR #870, opened 2019-08-27
fix reset zoom bug at sync line charts

HesamZamanpour:
#840 
reset zoom at sync line charts


junedchhipa:
Thank you for the fix.
Please make the change in `src/apexcharts.js` instead of `dist`

PR #942, opened 2019-09-29
Take into account value offset when calculating the height of radial bar - fixes #930

### LINKED ISSUE #930: Incorrect height for Radial Bar
Radial Bar height is sometimes calculated incorrectly, potentially cutting off values/labels. The height seems to be especially off when using `startAngle: -90`.
I've created a basic codepen that demonstrates the issue here:
https://codepen.io/aldis-ameriks/pen/qBWLxbK

I encountered the issue after bumping apexchart version to 3.8.6, the problem goes away after downgrading to 3.8.5.

aldis-ameriks:
I'm using apexcharts to render couple of semi-circle radial bars. After bumping version to 3.8.6, I noticed that the value texts get cut off because I'm rendering values slightly below the bar using positive `offsetY`. After a little digging, I noticed that the placelement of the value is not taken into account when calculating the height of the radial bar. It was not taken into account prior 3.8.6 version too, however, there seemed to be a constant spacing added for height, which was enough for my needs.

This PR adds functionality to include the value `offsetY` into the height calculation and sizing it appropriately, depending on the used angle. I'm not sure, but I suppose it may make sense to apply similar logic to other radial data labels.

Behaviour remains unchanged for full circle radial bars and if value offset is 0 or negative (thus shouldn't stretch the height downwards). And also when value text is not visible.

And of course, feel free to reject the PR if the change would cause other issues and/or is unwanted. Cheers!

Fixes #930 

## Type of change

## Checklist:

### DIFF
--- a/src/charts/Radial.js
+++ b/src/charts/Radial.js
@@ -94,7 +94,19 @@ class Radial extends Pie {
       )
     }
 
-    w.globals.radialSize = size - size / (360 / (360 - totalAngle))
+    const angleRatio = (360 - totalAngle) / 360
+
+    w.globals.radialSize = size - size * angleRatio
+
+    if (
+      this.radialDataLabels.value.show &&
+      this.radialDataLabels.value.offsetY > 0
+    ) {
+      const offset = this.radialDataLabels.name.show
+        ? this.radialDataLabels.value.offsetY + 16 // Additional offset when name label is rendered
+        : this.radialDataLabels.value.offsetY
+      w.globals.radialSize += offset * angleRatio
+    }
 
     elSeries.add(elG.g)
 


aldis-ameriks:
Before the changes and using `offsetY: 20` for the value
semi-circle-chart-before

After the changes and using `offsetY: 20` for the value
semi-circle-chart-after

aldis-ameriks:
For context, here is my use case how I'm rendering values below radial bar.
image

junedchhipa:
Hi @aldis-ameriks 
Thanks for your contribution. I pulled your code and tested it with other use-cases.
Unfortunately, it created some issues for other configuration. 
However, extending your code, I solved the issue through other ways.

Will be pushing the new fixes soon.

PR #961, opened 2019-10-05
fix reset zoom bug at sync line charts

HesamZamanpour:
#870 #840


HesamZamanpour:
sorry, because of wrong deleting previous branch i made that branch another time and changed `src/apexcharts.js` #870

PR #1128, opened 2019-12-11
Add filterSeries and updateAnnotations

ThinkSalat:
filterSeries: Allow filtering the series by passing a callback which receives the current series and returning the series you'd like the chart to use

updateAnnotations: passes the current annotations to a callback function allowing user to return annotations they wish to keep displayed. Currently if you have more than one type of annotation and want to remove only one, it's impossible.


ThinkSalat:
Can we get these features added? It barely changes anything and it would be very useful for your users.

src/apexcharts.js
```diff
@@ -706,15 +726,16 @@ export default class ApexCharts {
   revertDefaultAxisMinMax() {
     const w = this.w
 
-    w.config.xaxis.min = this.opts.xaxis.min || (Apex.xaxis && Apex.xaxis.min)
-    w.config.xaxis.max = this.opts.xaxis.max || (Apex.xaxis && Apex.xaxis.max)
+    w.config.xaxis.min = w.globals.lastXAxis.min
```
junedchhipa: @ThinkSalat This was old code and was reverted due to some issues. Can you please explain why this change is required?

ThinkSalat: Oh, I didn't touch that. I might have pulled down an older release or something?

junedchhipa: Yes, that's a possibility.
Can you please pull the latest master branch and send a new PR?

junedchhipa:
Closing due to inactivity.
Also, updateAnnotations will require more cleanup, hence removing and adding again is safest for now.

PR #1486, opened 2020-04-10
Fix for incorrect representation when using logarithmic scale

### LINKED ISSUE #1483: Incorrect representation when using logarithmic scale
# Bug report

When enabling logarithmic scale for the y-axis the data is not plotted correctly: the values of the logarithmic scale do not match the representation, they seem to be off by about a 10%.

Tried with various types of xaxis, all with the same result (categorey, numerical, datetime). If the logarithmic scale is switched off the representation is correct.

## Codepen

https://codepen.io/llmora/pen/ZEbEVmJ

## Explanation

**- What is the behavior you expect?**

The data points plotted should be aligned with the y-axis scale.

**- What is happening instead?**

The data points are plotted below the y-axis scale value (for instance for a value of 10 it is plotted at around the value of 9, a value of 100 is plotted around the value of 90, etc.)

**- What error message are you getting?**

No error message displayed or in the console. Tried going back all the way to v2.0.0

llmora:
# New Pull Request

When y-axis ticks are generated for logarithmic scales it does not take into account that ticks must be on power of 10 and above the max-y value. This results in data is not plotted correctly: the values of the logarithmic scale do not match the representation.

Fixes #1483

## Type of change

## Checklist:


junedchhipa:
What if the user needs to have a `min` value in log scale?

src/modules/Scales.js
```diff
@@ -152,45 +152,15 @@ export default class Range {
     }
   }
 
-  logarithmicScale(index, yMin, yMax, ticks) {
-    if (yMin < 0 || yMin === Number.MIN_VALUE) yMin = 0.01
+  logarithmicScale(yMax) {
+    let logs = []
```
PhilippMeissner, line 156:
```suggestion
    const logs = []
```

src/modules/Scales.js
```diff
@@ -152,45 +152,15 @@ export default class Range {
...
 
-    const base = 10
+    let ticks = Math.ceil(Math.log10(yMax)) + 1; // Get powers of 10 up to our max, and then one more
```
PhilippMeissner, line 158:
```suggestion
    const ticks = Math.ceil(Math.log10(yMax)) + 1; // Get powers of 10 up to our max, and then one more
```

PR #1489, opened 2020-04-11
Fix wrong polish translation of zoomOut

fracz:
"Oddaj" means "give back"
"Oddal" means "zoom out" :-)


junedchhipa:
Thank you for pointing this out!

PR #1545, opened 2020-04-24
Legend showForSingleSeries with no data issue fix

maxkochetov:
# New Pull Request

ApexCharts documentation states next:
```
showForSingleSeries: Boolean
Show legend even if there is just 1 series.
```
But in case when 
- no series data (empty array)
- and only 1 serie for the chart

The legend doesn't shows up.

I think that `showForSingleSeries = true` flag should be main indicator for legend visibility and overwrite any other conditions.

I also spend some time trying to add tests, but 

- for unit test i've checked that apexcharts project does not uses `toHaveBeenCalled`, and in this case the only good way to unit test is to check if `this.drawLegends()` was called inside Legend.js. so i decided to not add new ways to unit test

- as for e2e i didn't understand how to add sample for generation. if you can give me some reference or describe how to do that - i might add e2e test here, if needed

thanks :) 

## Type of change

## Checklist:

### DIFF
--- a/src/apexcharts.js
+++ b/src/apexcharts.js
@@ -128,10 +128,9 @@ export default class ApexCharts {
     gl.comboCharts = combo.comboCharts
     gl.comboBarCount = combo.comboBarCount
 
-    if (
-      ser.length === 0 ||
-      (ser.length === 1 && ser[0].data && ser[0].data.length === 0)
-    ) {
+    const allSeriesAreEmpty = ser.every(s => s.data && s.data.length === 0)
+
+    if (ser.length === 0 || allSeriesAreEmpty) {
       this.series.handleNoData()
     }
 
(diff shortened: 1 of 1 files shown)


src/apexcharts.js
```diff
@@ -128,10 +128,9 @@ export default class ApexCharts {
     gl.comboCharts = combo.comboCharts
     gl.comboBarCount = combo.comboBarCount
 
-    if (
-      ser.length === 0 ||
-      (ser.length === 1 && ser[0].data && ser[0].data.length === 0)
-    ) {
+    const allSeriesAreEmpty = ser.every(s => s.data && s.data.length === 0)
+
+    if (ser.length === 0 || allSeriesAreEmpty) {
```
maxkochetov, line 133: before it was checking only for 1 series for no data,
but better to check all passed series.data for an empty array to set an appropriate "No Data" state

PR #1626, opened 2020-05-21
Fixing old timestamp issue

danielbm:
# New Pull Request

The code was validating a timestamp by checking if the number of digits was equal to 13. However, dates before 08/08/2001 have a 12 digit timestamp, and other past and future dates might have other number of digits.

I changed the code to validate the timestamp by checking if the number is greater than zero.

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x ] My code follows the style guidelines of this project
- [ x ] I have performed a self-review of my own code
- [ x ] My changes generate no new warnings
- [ x ] New and existing unit tests pass locally with my changes


junedchhipa:
I would suggest adding `String(xlabels[j]).length !== 12 || String(xlabels[j]).length !== 13` will give a better hint to the user how long a timestamp should be. Because Unix timestamps are 10 characters.


Also, please remove the `dist` files as they are only produced when releasing a new build.
Thanks for your contribution!

danielbm:
I will write another pull request to address the issues!

PR #1712, opened 2020-06-27
Keep max an min axis in update options

gdenhez42:
# New Pull Request

We use react-apexcharts and got a similar problem than this one: https://github.com/apexcharts/react-apexcharts/issues/143
After some digging, the problem occur when we change the options, and after, we change the series. This is the fix I propose for this problem, please review carefully as I'm quite new to apexcharts. Thanks!

## Type of change

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### DIFF
--- a/dist/apexcharts.js
+++ b/dist/apexcharts.js
@@ -22893,8 +22893,8 @@
 
             if (overwriteInitialConfig) {
               // we need to forget the lastXAxis and lastYAxis is user forcefully overwriteInitialConfig. If we do not do this, and next time when user zooms the chart after setting yaxis.min/max or xaxis.min/max - the stored lastXAxis will never allow the chart to use the updated min/max by user.
-              w.globals.lastXAxis = [];
-              w.globals.lastYAxis = []; // After forgetting lastAxes, we need to restore the new config in initialConfig/initialSeries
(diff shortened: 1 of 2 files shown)


junedchhipa:
I have merged the PR as it looks fine to me.
Can you please give me a scenario where it was actually causing problems? Was it not updating the y-axis when you were calling `updateOptions`?
Can you create a codepen or something to illustrate?

gdenhez42:
Hello! Yes of course! I created this pen to illustrate the problem we have:
https://codepen.io/gdenhez/pen/WNrZvLZ

The chart is initialized with a min and a max xaxis. If you click to the "update series" button without clicking "update options", then the series gets updated and the min and max is still respected. If you click the "update options" which updates the min and max, it correctly update them. But then, if you click "update series", the new min and max is no longer respected (and even the old min and max is still not followed) 

We have a use case in which we can display data for a month, a week or a day range. When we change the range, it changes the options but then data for this range is fetched from our backend, and when it comes back, change the series only. The problem is then our range is not correctly displayed. Our workaround is currently to always change the id of our chart when our component update, which makes the react wrapper call updateOptions instead of updateSeries.

junedchhipa:
Thank you! The demo explained the bug.
👍

PR #1850, opened 2020-08-20
added scale property to chart configuration, by default is 1

### LINKED ISSUE #1849: Tooltips are not correct when chart is scaled with transform
# Bug report

## Codepen

https://codepen.io/linkaynn/pen/RwaGvVP

## Explanation

When you scale a chart the tooltip position and information is not correct, when scale is `> 1` it takes the points at the right from mouse position and if scale is `< 1` takes the left points

**What is the behavior you expect?**

That tooltips takes correct positions and information

Linkaynn:
# New Pull Request

Now charts has the possibility to scale with the css transform

Fixes #1849 

## Type of change

- [X] This change requires a documentation update

## Checklist:

### DIFF
--- a/src/modules/settings/Options.js
+++ b/src/modules/settings/Options.js
@@ -291,6 +291,7 @@ export default class Options {
         group: undefined,
         offsetX: 0,
         offsetY: 0,
+        scale: 1,
         selection: {
           enabled: false,
           type: 'x',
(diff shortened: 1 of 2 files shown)


junedchhipa:
Thanks for the quick PR 😀 
So, the chart scaling has to be done in CSS and property has to be set in JS options? Am I right?
How about scaling the chart in JS itself when we already have the chart.scale property?

Linkaynn:
You'r welcome! :smile:

Right now the only way to solve this limitiation without looping all the DOM is making the scalation in parallel, when CSS changes, JS must change manually.

Maybe this is not the perfect solution, but it really solve the point that chart shows correctly the tooltips because it some parent container has a scale (CSS or JS).

In case that you scale the chart by JS has the same point that scale in CSS in this case.

PR #1999, opened 2020-10-26
removes unnecessary update when window height changes

FunkiR:
# New Pull Request

Hi, I am using graphics inside an iframe that dynamically changes its height. This triggers the resize event and all charts start updating. Is such a solution to this problem possible?

## Type of change

## Checklist:

### DIFF
--- a/src/apexcharts.js
+++ b/src/apexcharts.js
@@ -37,14 +37,6 @@ export default class ApexCharts {
     initCtx.initModules()
 
     this.create = Utils.bind(this.create, this)
-
-    let width = window.innerWidth
-    this.windowResizeHandler = () => {
-      if (width !== window.innerWidth) {
-        width = window.innerWidth
-        this._windowResize.call(this)
-      }
-    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


junedchhipa:
There is a property `chart.redrawOnParentResize` which can be turned off if you don't want the charts to be updated on parent window resize. Try and let me know if it works.

FunkiR:
No, this property is used in _parentResizeCallback, but I need to remove the height triggering in windowResizeHandler.
```javascript
window.addEventListener('resize', this.windowResizeHandler)
window.addResizeListener(
     this.el.parentNode,
     this._parentResizeCallback.bind(this)
)
```

junedchhipa:
Ok, can you test if your code doesn't turn off the actual behavior?
Try placing the chart in a container, and only resize that parent (not window).
If possible, please show a GIF that does that.

FunkiR:
Everything works, but I don't like my first solution) Could it be better to add the redrawOnWindowResize property? Here is a [example](https://codepen.io/funkir/pen/JjKrLZq)

junedchhipa:
Thanks for providing an alternative.
This looks better. I will merge soon.

PR #2296, opened 2021-03-17
REF: Do not hardcode rangebar tooltip

venarius:
Hey,
currently the "RangeBar" tooltip has hardcoded HTML elements in it that show the range of the bar in numbers. A user cannot change these outputs as they cannot be overridden at the moment.

This change removes the hardcoded start and end values from the tooltip of the "RangeBar" and instead gives them as options to the y-label tooltip formatter to allow users to customize the output.

## Type of change

- [x] Refactor

### DIFF
--- a/src/charts/RangeBar.js
+++ b/src/charts/RangeBar.js
@@ -371,7 +371,7 @@ class RangeBar extends Bar {
       }
 
       if (typeof yLbFormatter === 'function') {
-        ylabel = yLbFormatter(ylabel, { ...opts, ...{ start, end } })
+        ylabel = yLbFormatter(ylabel, opts)
       }
     }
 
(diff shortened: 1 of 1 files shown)


junedchhipa:
Thanks for the idea. I have made the change in this commit - https://github.com/apexcharts/apexcharts.js/commit/700f6612b240c8daef6559449fb18c2140858027
I am not accepting this PR because it removes the default tooltip. Instead, I have kept start and end in opts parameter.

venarius:
> I am not accepting this PR because it removes the default tooltip. Instead, I have kept start and end in opts parameter.

But this sadly doesn't fix the issue here. The PR does not remove the default tooltip. It only removes the start and end numbers which are currently being displayed next to the label. User are not able to turn them off at the moment as they are hardcoded into the tooltip right now.

I attached a screenshot for better description of the problem we are facing.

screen

PR #2310, opened 2021-03-24
issue-1959: make sure e gets default values if e is null

### LINKED ISSUE #1959: Unable to create Apexcharts inside Angular material tabs when using Firefox
# Bug report

## Stackblitz

https://stackblitz.com/edit/angular-xk6h8g?file=src/app/chartconfig/apexchart-config.ts

## Explanation
When using Firefox, Apexcharts in Angular 10 is not working when placed in a `<mat-tab></mat-tab>`.
This works flawlessly when using a Chromium-based browser, but fails when using Mozilla Firefox.

Charts to be working when placed under Angular Material Tabs.

- What is happening instead?
When opened in Firefox, an error is displayed in debugger, seemingly when trying to initialize charts on tabs that are not currently open. All Apexcharts functionality stops, after only the chart on the currently open tab has been drawn.

In Visual Code debugger console:
> columnNumber:39
> fileName:"http://localhost:4200/polyfills.js"
> lineNumber:3006
> message:"Uncaught (in promise): TypeError: e is null

ch0001:
# New Pull Request
Change in dist/apexcharts.common.js  at  `a.Matrix`. 

In Firefox the value of `e` can be `null`. And because `t(null) === "object"` , `e` doesn't get default values but keeps being `null`.

```
create: function(e) {
     var i = p([1, 0, 0, 1, 0, 0]);
     e = e === null ? i :   //<-- additional test
           e instanceof a.Element ?  e.matrixify() : 
           "string" == typeof e ?   p(e.split(a.regex.delimiter).map(parseFloat)) : 
           6 == arguments.length ?  p([].slice.call(arguments)) : 
           Array.isArray(e) ? p(e) : 
           "object" === t(e) ? e : i;
     for (var s = v.length - 1; s >= 0; --s) { this[v[s]] = null != e[v[s]] ? e[v[s]] : i[v[s]] }
}
```

Fixes #1959

## Type of change

## Checklist:


junedchhipa:
Thanks for the contribution, I have added a check as per your suggestion.

PR #2332, opened 2021-04-01
Use "module" in package.json for ES build

vojty:


junedchhipa:
Sorry, the esm build currently has issues with angular/react.
Many people reported bugs when loading esm file, hence I removed the esm build for now.

PR #2404, opened 2021-05-13
Change dataURI method to support setting resolution on PNG export

### LINKED ISSUE #1591: Higher resolution on PNG export
While my graphs look nice on the webpage, when I download them to put them in a presentation or similar, they have a rather low resolution.

It would be awesome to be able to define a width/height or scale factor for PNG export so I could easily download the same graph with a better resolution.

**Describe alternatives you've considered**	
a) I have tried to just make the graph bigger: This does not work well as I also would need to increase the fonts but then still the lines, legend and other things are really small. Seems like a lot of fiddling around to get close to a better resolution.
b) adding `transform="scale(2)"` to the svg element seems to zoom the svg fine. I just don't know how to shortly do this during the export.
c) another option might be to create a second invisible bigger export canvas, copy the canvas content there and scale it and export that canvas. This would also allow to add other features like "add a watermark only on export".

**Additional context**	
I believe exports with a higher resolution would be very useful for many users. I might even be able to find some sponsoring for this feature.

rezendeneto:
# New Pull Request

Change the `dataURI` method to optionally pass the `options` object as argument, through which it is now possible to set either a `scale` or a `width` in order to scale the exported PNG file.

Fixes #1591

## Type of change

## Checklist:


junedchhipa:
Thanks for creating a PR.
Can you please remove the `dist` files. Those files should only be updated on a new release.

rezendeneto:
Hi @junedchhipa, thanks for your reply - I've just removed the `dist` files as you suggested.

junedchhipa (approved):

rezendeneto:
@junedchhipa thanks for merging! Do you plan to release a new version soon?

junedchhipa:
in a couple of days

vaspadi:
When will this be added to the vue component?

cmfc31:
Looking forward for the implementation in the vue component as well

a-shafaat:
@vaspadi @cmfc31 
I had the same question and found a way to do it: **use [exec](https://apexcharts.com/docs/methods/#exec)**
this.$apexcharts.exec('chart-id', 'dataURI', { width: 1000 })).imgURI

Ali-Zaraket:
Hi, is this supported for reactjs library, where I can set the resolution in options ?

PR #2614, opened 2021-09-13
Added support for series specific border radius on bar chart

### LINKED ISSUE #2592: borderRadius is not working properly with type "Bar with Negative Values"
### Description

If you follow [this exemple](https://apexcharts.com/javascript-chart-demos/bar-charts/bar-with-negative-values/) and you try to add a border raduis, it won't work properly. The border radius is only applied to the left bars and not to the right bars.

My code : 
```js
  var options = {
    series: [
      {
        name: 'Males',
        data: [0.4, 0.65, 0.76, 0.88, 1.5, 2.1, 2.9, 3.8, 3.9, 4.2, 4, 4.3, 4.1, 4.2, 4.5,
          3.9, 3.5, 3
        ]
      },
      {
        name: 'Females',
        data: [-0.8, -1.05, -1.06, -1.18, -1.4, -2.2, -2.85, -3.7, -3.96, -4.22, -4.3, -4.4,
          -4.1, -4, -4.1, -3.4, -3.1, -2.8
        ]
      }
    ],
    chart: {
      type: 'bar',
      height: 440,
      stacked: true
    },
    colors: ['#008FFB', '#FF4560'],
    plotOptions: {
      bar: {
        horizontal: true,
        barHeight: '80%',
        borderRadius: 5   // Here is the issue ...
      },
    },
    dataLabels: {
      enabled: false
    },
    stroke: {
      width: 1,
      colors: ["#fff"]
    },
    grid: {
      xaxis: {
        lines: {
          show: false
        }
      }
    },
    yaxis: {
      min: -5,
      max: 5,
    },
    xaxis: {
      categories: ['85+', '80-84', '75-79', '70-74', '65-69', '60-64', '55-59', '50-54',
        '45-49', '40-44', '35-39', '30-34', '25-29', '20-24', '15-19', '10-14', '5-9',
        '0-4'
      ],
      title: {
        text: 'Percent'
      },
      labels: {
        formatter: function (val) {
          return Math.abs(Math.round(val)) + "%"
        }
      }
    },
  };

  chart.render();
```

### Steps to Reproduce

1. Try the code below

### Expected Behavior

All bars have rounded corners.

### Actual Behavior

Only one side of the chart has bars with rounded corners.

### Reproduction Link

https://codepen.io/plapointe6/pen/QWgEKwJ

brianlagunas:
# New Pull Request

`plotOptions.bar.borderRadius` now excepts both a number and a number array.

This PR is a combination of a bug fix and a new feature.  By allowing `plotOptions.bar.borderRadius` to accept both a `number` and a `number[]`, developers can now apply a border radius to a specific series in their chart.

Fixes #2592

The `borderRadius` property now behaves as follows:

`borderRadius: 5` - when setting the border radius to a number, there is **no change in behavior**. All series will receive the same border radius value. Unless it's a stacked bar chart. In that case all series would have a border radius of 0 except for the last series in the stacked bar chart.

`borderRadius: [5, 10, 5, 3]` - when setting the border radius to an array, the border radius value will be applied to each series according to the index of the series. In this example, if there are 4 series, then the border radius would be applied in the order in which it is defined. `Series 0 = Radius 5`, `Series 1 = Radius 10`, `Series 2 = Radius 5`, `Series 3 = Radius 3`. This is specifically important when dealing with a stacked series bar chart. By default, all series in a stacked series bar chart would be forced to have a radius of 0 except for the last series index which would then have a border radius value set. With this new feature, this will force all series, even a stacked bar series, to have a border radius set.

`borderRadius: [5]` - when setting the border radius to a array with a single value, then all series will receive the same border radius value. This is specifically important when dealing with a stacked series bar chart. By default, all series in a stacked series bar chart would be forced to have a radius of 0 except for the last series index which would then have a border radius value set. With this new feature, this will force all series, even a stacked bar series, to have a border radius set.

In the case where there are more series than border radius values in the array, the last value in the borderRadius array will be applied to all remaining series.

## Type of change

## Checklist:

### DIFF
--- a/src/charts/common/bar/Helpers.js
+++ b/src/charts/common/bar/Helpers.js
@@ -381,12 +381,23 @@ export default class Helpers {
    **/
   getRoundedBars(w, opts, series, i, j) {
     let graphics = new Graphics(this.barCtx.ctx)
-    let radius = w.config.plotOptions.bar.borderRadius
+    let radius = 0
+
+    const borderRadius = w.config.plotOptions.bar.borderRadius
+    const borderRadiusIsArray = Array.isArray(borderRadius)
+    if (borderRadiusIsArray) {
+      const radiusIndex =
+        i > borderRadius.length - 1 ? borderRadius.length - 1 : i
+      radius = borderRadius[radiusIndex]
+    } else {
+      radius = borderRadius
+    }
 
     if (
       w.config.chart.stacked &&
       series.length > 1 &&
-      i !== this.barCtx.radiusOnSeriesNumber
+      i !== this.barCtx.radiusOnSeriesNumber &&
+      !borderRadiusIsArray
     ) {
       radius = 0
     }
--- a/types/apexcharts.d.ts
+++ b/types/apexcharts.d.ts
@@ -487,7 +487,7 @@ type ApexPlotOptions = {
     columnWidth?: string
     barHeight?: string
     distributed?: boolean
-    borderRadius?: number
+    borderRadius?: number | number[]
     rangeBarOverlap?: boolean
     rangeBarGroupRows?: boolean
     colors?: {


junedchhipa:
Looks good!
1 thing we need to check is what happens when the user toggles the series using legend? Does it show any weird behavior there? Merging this.

junedchhipa (approved):

junedchhipa:
I see you have updated the TypeScript definitions as well 👍🏻 
One place where it will need to be also updated is the `ng-apexcharts` library.

For future reference, any change that goes to the ApexCharts TypeScript definitions also has to be done at https://github.com/apexcharts/ng-apexcharts/blob/master/projects/ng-apexcharts/src/lib/model/apex-types.ts

brianlagunas:
I checked the toggling of the legend and the feature works with no issues.

filippolegati:
If I have a stacked chart with some dynamic data (e.g.  first data[10,10,2,2,3]  and second data[2,4,**0,0,**2]), the third and fourth column wont have the border radius effect on them because the last item of the column is not present. Is there a way to set it with the array?

ak-ragnor:
`var optionstack = {
							series: [{
								name: 'Current Sales Value',
								data: data.greenBar
							},
							{	
								name: 'Shortfall Value',
								data: data.greyBar
							}],
							colors : ['#38AB0F', '#CCC'],
							chart: {
								type: 'bar',
								height: 350,
								width:'100%',
								stacked: true,
							},
							noData: {
							    text: "No data text",
							    align: "center",
							    verticalAlign: "middle",
							  },
							responsive: [{
								breakpoint: 375,	
								options: {
									legend: {
										position: 'bottom',
										offsetX: -10,
										offsetY: 0
									}
								}
							}],
							plotOptions: {
								bar: {
									horizontal: false,
									borderRadius: [5, 5],
									barWidth: '60%',
									columnWidth:'40%',
									borderRadiusOnAllStackedSeries: true
								},
							},
							xaxis: {
								categories: data.barCategories
							},
							yaxis: {
								tickAmount: 5,
								max: data.yaxisMax
							},
							dataLabels: {
						          enabled: false
						        },
							legend: {
								position: 'bottom',
								horizontalAlign: 'center',
							},
							fill: {
								opacity: 1
							},
							tooltip: {
								enabled: true,
								shared: true,
								intersect: false,
							    fixed: {
							          enabled: false,
							          position: 'topRight',
							          offsetX: 0,
							          offsetY: 0,
							    },
							},
							states: {
								  hover: {
								   filter: {
								      type: 'none'
								    }
								  }
								}
						};

						var stackedChart = new ApexCharts(document.querySelector("#stackedChart"), optionstack);
						stackedChart.render();`
						
						
i have been trying to set border radius but still this is not working

PR #2647, opened 2021-09-20
Implemented discrete markers for scatter

### LINKED ISSUE #2639: markers.discrete option is not working when using multiple scatter series
### Description
I'm visualizing multiple series of scattered data on a line chart and I would like to customize one of the dots. 
[markers.discrete](https://apexcharts.com/docs/options/markers/) sounds like a perfect candidate to implement this but unfortunately I found it to not work with multiple series of scatter type.

### Steps to Reproduce

1. Set up a basic chart of type 'line'
2. Add two series of type 'scatter'
3. Add markers.discrete option and modify one of the data points to alter its appearance.

### Expected Behavior
Dots should get customized according to the options passed in markers.discrete.

### Actual Behavior
No styles are applied to the specified dot.

### Screenshots
Screenshot 2021-09-17 at 16 28 18

### Reproduction Link
[Codepen](https://codepen.io/gruszkakarol/pen/NWgydpr)

brianlagunas:
# New Pull Request

Implemented discrete markers for scatter.

Fixes #2639

## Type of change

## Checklist:

### DIFF
--- a/src/charts/Scatter.js
+++ b/src/charts/Scatter.js
@@ -1,5 +1,4 @@
 import Animations from '../modules/Animations'
-import Fill from '../modules/Fill'
 import Filters from '../modules/Filters'
 import Graphics from '../modules/Graphics'
 import Markers from '../modules/Markers'
--- a/src/charts/Scatter.js
+++ b/src/charts/Scatter.js
@@ -102,18 +102,15 @@ export default class Scatter {
     let i = realIndex
     let anim = new Animations(this.ctx)
     let filters = new Filters(this.ctx)
-    let fill = new Fill(this.ctx)
     let markers = new Markers(this.ctx)
     const graphics = new Graphics(this.ctx)
 
-    const markerConfig = markers.getMarkerConfig('apexcharts-marker', i)
-
-    let pathFillCircle = fill.fillPath({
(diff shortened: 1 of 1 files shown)


junedchhipa:
We will need to keep the `pathFillCircle` as it is used for advanced fill options (like gradient/pattern in bubble charts)
If you open the 3d-bubble demo in `samples/vanilla-js/bubble/3d-bubble.html`, you will see it doesn't render anything after this change.

brianlagunas:
So what kind of logic can I use to determine if we need to apply the pathFillCircle or the markerConfig Fill?

junedchhipa:
Yeah, I am looking at the code, and I probably made a mistake when combining both.
markerConfig is used to get certain properties (height/width/radius/shape/color), while pathFillCircle is only used for the fill option.
When we call `fillPath` function, we can also send `color` in the options so that color gets applied.
Let me also test it first

junedchhipa:
@brianlagunas Your code helped me to dig the root of the cause. However, I will open a new PR because I had to refactor some code and do some cleanup to the older code.

brianlagunas:
Sounds good! I will close this PR and wait to review yours so I can better learn the code base.

Thanks!

PR #2755, opened 2021-11-08
Reduce innerHTML usage to improve performance and reduce XSS risk

### LINKED ISSUE #816: xaxis:categories are open to XSS attack. Tooltip does not escape the category but bottom of the graph does it
I am using react but I have been able to reproduce the issue on the suggested codepen.io (https://codepen.io/apexcharts/pen/bxzgZJ). I have fordked it into https://codepen.io/ismamai/pen/LwWJoa

Here the issue is that the categories are escaped when shown on the bottom of the table but they are not on the tooltip. I tried escaping them my self, but then they get escaped twice on the bottom

## Explanation

the  '<'img src="http://tiny.cc/p3vu9y"> should be escaped and shown as text on the tooltip.
- What is happening instead?
Image is being displayed
Input data from client could contain any xss vector attack. We normally escape the data to avoid xss attacks, but here we are not sure about how to proceed

illandril:
# New Pull Request

Changes several `.innerHTML` calls with `.textContent` and `createElement`/`appendChild` calls.
This both sanitizes the input to protect against XSS attacks (like #816) and drastically improves the performance (especially around tooltips).

Fixes #816 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

These changes will break anyone who is currently relying on this unsafe behavior and passing in raw HTML to any of these values.

## Checklist:

Unfortunately, the e2e tests don't pass on my system in the unmodified master code (before any of my changes). Since an e2e test would be the appropriate place to test these, and since I can't get the current ones to pass on my system, I did not write new tests. I did manually test the changes, and the e2e tests fail in identical ways both before and after my changes.

These changes do not fix all XSS and performance risks around using innerHTML - Toolbar, PointAnnotations, Legend, and svg all use innerHTML, but they all would need a more substantial update (with a higher risk of impacting more current users) to not rely on innerHTML.

### DIFF
--- a/src/modules/Toolbar.js
+++ b/src/modules/Toolbar.js
@@ -190,7 +190,7 @@ export default class Toolbar {
     }
     for (let i = 0; i < menuItems.length; i++) {
       this.elMenuItems.push(document.createElement('div'))
-      this.elMenuItems[i].innerHTML = menuItems[i].title
+      this.elMenuItems[i].textContent = menuItems[i].title
       Graphics.setAttrs(this.elMenuItems[i], {
         class: `apexcharts-menu-item ${menuItems[i].name}`,
         title: menuItems[i].title
--- a/src/modules/legend/Legend.js
+++ b/src/modules/legend/Legend.js
@@ -199,7 +199,7 @@ class Legend {
 
       let elLegendText = document.createElement('span')
       elLegendText.classList.add('apexcharts-legend-text')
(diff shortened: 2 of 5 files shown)


aavezel:
I think this change is harmful because it could break existing usecases. For example, when using html to format part of the legend (bold, italics). IMHO, it is better to control the security of charts display at the stage of chart setup, not in the same chart

PR #2767, opened 2021-11-10
Inject ApexCharts CSS to shadow root if we're in a shadow DOM context

### LINKED ISSUE #238: ShadowDOM support
The apexcharts js file adds CSS styles into the document's header. This makes the graph break when the chart is used in a webcomponent as the styles never reach the DOM nodes because of shadowDOM.

I think the CSS styles should either exposed as a file in the dist directory or add them to the shadowRoot node OR to the document's header

Cheers,
Pupix

LeaVerou:
# New Pull Request

Previously, all ApexCharts CSS was injected in the `<head>` of the document ApexCharts was included in, through [StyleInject](https://github.com/egoist/style-inject/blob/master/src/index.js) that was being called by [`rollup-plugin-postcss`](https://github.com/egoist/rollup-plugin-postcss).

This PR disables the automatic injection, and instead injects it manually. We first check if we're in a Shadow DOM context, and append the CSS to the shadow root if so. Otherwise, we check if the CSS has already been added to the current document, and insert it if not.

While the primary goal of this PR was to fix the Shadow DOM problem, it introduces a few other improvements:
- Since the injection is now context-aware, even for non-shadow contexts, we inject in the chart element's document, not the document the ApexCharts library was included on. While in most cases these are the same, they are not when the library is included in one document, and the chart is in another (e.g. an iframe).
- With this change, if ApexCharts are included, but no charts are actually drawn on the current page, this won't inject any CSS, whereas previously just including the Apex Charts module produced side effects (injected CSS).
- I also added an `Util.is()` helper to perform type-checking that works across different window objects, and applied it on existing code that was already using this method without a helper.
- The two unrelated changes in Utils.js (L7, L36-37 and L226-228) are to apply the current Prettier rules (I guess code was changed there but Prettier was not run).

Fixes #238

## Type of change

## Checklist:

### DIFF
--- a/build/config.js
+++ b/build/config.js
@@ -66,6 +66,7 @@ function rollupConfig(opts) {
         preferConst: true
       }),
       postcss({
+        inject: false,
         plugins: []
       }),
       svgo({
--- a/src/apexcharts.js
+++ b/src/apexcharts.js
@@ -13,6 +13,7 @@ import YAxis from './modules/axes/YAxis'
 import InitCtxVariables from './modules/helpers/InitCtxVariables'
 import Destroy from './modules/helpers/Destroy'
 import { addResizeListener, removeResizeListener } from './utils/Resize'
+import apexCSS from './assets/apexcharts.css'
 
 /**
  *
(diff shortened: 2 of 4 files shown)


junedchhipa (approved):
Approved!

PR #2983, opened 2022-02-17
feat: Add sub-groups in xAxis labels

DavidMarquezF:
# New Pull Request

Added sub-groups for xAxis. This allows for 1 level of groupping in the xAxis and is something offered by other apps like excel.

To achieve this the model is the following: 

```
 {
    chart: {
        type: 'bar',
    },
    series: [{
        name: 'sales',
        data: [30, 40, 45, 50, 49, 60, 70, 91, 125, 10]
    }],
    xaxis: {
        categories: [12, 1992, 1993, 1994, 1995, 1996, 1997, 1998, 1999, 10],
        group: {
            groups: [{ title: "Test", cols: 3 }, { title: "Test1", cols: 2 }, { title: "Test2", cols: 4 },],
        },
        title: {
            text: "AAA"
        }
    }
}
```

The fact that you can specify the number of columns each group takes gives a lot of flexibility to do any combination you want

4mcy30p1

## Type of change

## Checklist:

### DIFF
--- a/samples/source/bar/bar-with-group-labels.xml
+++ b/samples/source/bar/bar-with-group-labels.xml
@@ -1,0 +1,63 @@
+<title>Bar with Group Labels</title>
+<chart>
+    <options>
+chart: {
+  type: 'bar',
+  height: 380
+},
+plotOptions: {
+  bar: {
+    barHeight: '100%',
+    distributed: true,
+    horizontal: true,
+    dataLabels: {
+      position: 'bottom'
+    },
+  }
+},
+colors: ['#33b2df', '#546E7A', '#d4526e', '#13d8aa', '#A5978B', '#2b908f', '#f9a3a4', '#90ee7e',
+  '#f48024', '#69d2e7'
+],
+xaxis: {
+  categories: ['South Korea', 'Canada', 'United Kingdom', 'Netherlands', 'Italy', 'France', 'Japan',
+    'United States', 'China', 'India'
+  ],
+  group: {
+      groups: [{ title: "Test", cols: 3 }, { title: "Test1", cols: 2 }, { title: "Test2", cols: 4 }]
+  }
+},
+yaxis: {
+  labels: {
+    show: false
+  }
+},
+title: {
+    text: 'Custom DataLabels',
+    align: 'center',
+    floating: true
+},
+subtitle: {
+    text: 'Category Names as DataLabels inside bars',
+    align: 'center',
+},
+tooltip: {
+  theme: 'dark',
+  x: {
+    show: false
+  },
+  y: {
+    title: {
+      formatter: function () {
+        return ''
+      }
+    }
+  }
+}
+</options>
+
+    <series>
+[{
+  data: [400, 430, 448, 470, 540, 580, 690, 1100, 1200, 1380]
+}]
+</series>
+</chart>
--- a/samples/vanilla-js/bar/bar-with-group-labels.html
+++ b/samples/vanilla-js/bar/bar-with-group-labels.html
@@ -1,0 +1,123 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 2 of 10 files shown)


DavidMarquezF:
As of now this has some limitations which should be fixed before merging, but since I wasn't sure if this feature would be integrated in the library I decided to not do them for now. If there is an interest to move this forward I'll try to implement the rest of the things.


- [ ] Cannot specify formatters, colors, type and other properties for the group. A decision should be made weather this is worth it or not
- [ ] I don't calculate isXNumeric for the groups, for now I just use the existing isXNumeric, which is wrong
- [ ] Categories only support the ticks when `tickPlacement` is `between` and the `tickAmount` is not set or set to `datapoints`
- [ ] Create tests?

Moreover I did a modification which I would like someone to check if it was possible. In order to not repeat code for the labels and the group labels I created a function reusing most of it. Prior to this modification there were two variables: `this.drawnLabels` and `this.drawnLabelsRects` which were scoped to `this` and not the function. Now they are scoped at the function level because I didn't see any other usage for them. However, maybe it was being done for a reason that I haven't been able to guess.

DavidMarquezF:
Thanks for your response @junedchhipa.

I'm seeing that with these new additions I have to change to logic of the calculation of the xaxisHeight in `getxAxisLabelsCoords`, since if the fontsize is different for the group labels the height will have to vary accordingly. 

Does it make sense to change the calculation in order to also take into account the groups? Or would it be better to create a separate function to calculate the groups heigth (which should be simpler because there will not be timescales, formatters, etc.) and then also pass this height to `conditionalChecksForAxisCoords` in order to calculate `xAxisHeight`?

DavidMarquezF:
> formatter is not required for the groups as the values won't be dynamic, as I guess groups will be user-supplied. But colors and types options should be provided.

With this you mean that I should only allow for the same options that `xaxis.labels.style offers`? Or also things like `xaxis.labels.(show, trim, rotate, etc.)`? Adding styles should be quite straight forward, but some other features like rotate could be a bit more complicated to implement (or at least that's the impression I got by looking around the src code)

DavidMarquezF:
The tests are not passing, but they also do not pass on the master branch. I made an issue about it #2993. I also think that if the e2e tests pass it should mean that this addition will most probably not cause any issues no?

junedchhipa:
> Does it make sense to change the calculation in order to also take into account the groups? Or would it be better to create a separate function to calculate the groups heigth (which should be simpler because there will not be timescales, formatters, etc.) and then also pass this height to `conditionalChecksForAxisCoords` in order to calculate `xAxisHeight`?

In my opinion, it is better to create a new function to calculate the groups height and pass this height to other places.

junedchhipa:
> With this you mean that I should only allow for the same options that xaxis.labels.style offers? Or also things like xaxis.labels.(show, trim, rotate, etc.)? Adding styles should be quite straight forward, but some other features like rotate could be a bit more complicated to implement (or at least that's the impression I got by looking around the src code)

Only the style options. Below is the list
```
 style: {
  colors: [],
  fontSize: '12px',
  fontWeight: 400,
  fontFamily: undefined,
}
```

junedchhipa:
Ignore the e2e tests, for now, I will create some tests to cover this.

DavidMarquezF:
Okay perfect! Styles are already implemented. By default the group labels inherit the styles from the xaxis labels style property. The styles property in group will allow to overwrite these

I will work on the height calculation tomorrow probably

DavidMarquezF:
Okay, I was able to finish the proposed changes:
4uth369q

Big fontsize:
yg6df76u


Single color in group colors
voe2ggys

Array of colors in group colors:
srj9akgu

junedchhipa:
@DavidMarquezF After testing a couple of different charts, I feel the groups shouldn't be rotated when the x-axis labels are rotated. They look dissolved into the primary x-axis labels and there isn't a need to rotate group labels because they will be less in numbers than the primary x-axis labels.

Can you please modify the group labels not to rotate them even though the x-axis labels are rotated?

junedchhipa:
I have reverted this PR, so please open a new PR

DavidMarquezF:
Yes I will do it when I have time. So they should never rotate?

PR #2986, opened 2022-02-18
fix(ts): update a few types

innocenzi:
This PR fixes a few of the types to be more in-line with the actual code.

## Checklist:

### DIFF
--- a/types/apexcharts.d.ts
+++ b/types/apexcharts.d.ts
@@ -40,8 +40,9 @@ declare class ApexCharts {
   addPointAnnotation(options: any, pushToMemory?: boolean, context?: any): void
   removeAnnotation(id: string, options?: any): void
   clearAnnotations(options?: any): void
-  dataURI(options?: { scale?: number, width?: number }): Promise<void>
+  dataURI(options?: { scale?: number, width?: number }): Promise<{ imgURI: string }>
   static exec(chartID: string, fn: string, ...args: Array<any>): any
+  static getChartByID(chartID: string): ApexCharts|undefined
   static initOnLoad(): void
 }
 


types/apexcharts.d.ts
```diff
@@ -40,9 +40,20 @@ declare class ApexCharts {
   addPointAnnotation(options: any, pushToMemory?: boolean, context?: any): void
   removeAnnotation(id: string, options?: any): void
   clearAnnotations(options?: any): void
-  dataURI(options?: { scale?: number, width?: number }): Promise<void>
+  dataURI(options?: { scale?: number, width?: number }): Promise<{ imgURI: string }>
```
junedchhipa, line 43: dataURI can also return `blob`

```
if (canvas.msToBlob) {
  // IE and Edge can't navigate to data urls, so we return the blob instead
  let blob = canvas.msToBlob()
  resolve({ blob })
} else {
  let imgURI = canvas.toDataURL('image/png')
  resolve({ imgURI })
}
```

types/apexcharts.d.ts
```diff
@@ -40,9 +40,20 @@ declare class ApexCharts {
...
   static exec(chartID: string, fn: string, ...args: Array<any>): any
+  static getChartByID(chartID: string): ApexCharts|undefined
   static initOnLoad(): void
+  exports: {
```
junedchhipa, line 47: How is this `exports` object relevant here? It is not publicly exposed in ApexCharts API. How are you using it?

junedchhipa (commented):
Please explain how this new `exports` object is going to help by adding them to types definitions.

types/apexcharts.d.ts
innocenzi: Even though it's not documented or anything, it's still exposed. In our case, we use it to perform diverse operations that are implemented but hardly accessible from userland.

We have a custom toolbar, which we want to expose SVG/PNG/CSV download buttons, among other things. So we use `instance.exports.exportToSVG()`, `instance.exports.triggerDownload()` and other methods directly - where `instance` is an ApexChart instance retrieved by `getChartByID`.

junedchhipa (changes requested):

junedchhipa (approved):

PR #3078, opened 2022-03-31
Fix #2964 - Stacked Area chart: Markers of data points disordered and…

### LINKED ISSUE #2964: [3.33.2] Stacked Area chart: Markers of data points disordered and at the wrong layer in Firefox
### Description

The issue does not exist in Chrome 98.0.4758.82 (Official Build) (64-bit).

Reproduced in Firefox:
- "91.3"
- 91.5.1esr (64-Bit)
- 91.6.0esr (64-Bit)

### Steps to Reproduce

Codepen:

1. open in Chrome
=> okay
2. mouse-over the first data point
=> okay
3. repeat steps 1, 2 in Firefox
=> black marker will be on white curve, white marker will be on black curve
=> right half of white marker will be hidden by diagram
4. mouse-over the second data point
=> white marker will be completely hidden by diagram

MVE:

1. checkout
2. `npm run build`
3. `npm start`
4. open http://localhost:3000 in Chrome
=> okay
5. mouse-over the first data point
=> okay
6. repeat steps 4, 5 in Firefox
=> black marker will be on white curve, white marker will be on black curve
=> right half of white marker will be hidden by diagram
7. mouse-over the second data point
=> white marker will be completely hidden by diagram

### Expected Behavior

Markers on mouse-over behave in Firefox as they behave in Chrome.

### Actual Behavior

Markers displayed on mouse-over have the wrong order and some of them are layered behind the diagram instead of in front of it.

### Screenshots

Firefox_Markers_MisorderedAndOnWrongLayer_0
Firefox_Markers_MisorderedAndOnWrongLayer_1

### Reproduction Link

- [Codepen](https://codepen.io/zsar/pen/xxPdPzY)
- [MVE for local PC](https://github.com/Zsar/MVE_ApexCharts_Firefox)

sbke-mms:
… at the wrong layer in Firefox

# New Pull Request

This fix orders markers in stacked area charts properly in both Firefox and Chrome. This was tested with the MVE from @Zsar (I also tried it for unstacked and with more series entries) and the project where we noticed the issue. However I am uncertain on the effect on other charts so this should be checked as well.

Fixes #2964

## Type of change

## Checklist:

### DIFF
--- a/src/modules/tooltip/Utils.js
+++ b/src/modules/tooltip/Utils.js
@@ -297,8 +297,11 @@ export default class Utils {
     markersWraps.sort((a, b) => {
       return Number(b.getAttribute('data:realIndex')) <
         Number(a.getAttribute('data:realIndex'))
-        ? 0
-        : -1
+        ? 1
+        : Number(b.getAttribute('data:realIndex')) >
+          Number(a.getAttribute('data:realIndex'))
+        ? -1
+        : 0
     })
 
     let markers = []


sbke-mms:
I first suspected `drawDynamicPoints` in `Marker.js` to be the cause for this since the builds before a38e296 didn't have the issue but changes there didn't seem to affect the chart. Maybe that should be checked as well

src/modules/tooltip/Utils.js
```diff
@@ -297,8 +297,11 @@ export default class Utils {
     markersWraps.sort((a, b) => {
       return Number(b.getAttribute('data:realIndex')) <
```
Zsar, line 298: As we already use a code block, it might be sensible to extract the `getAttribute` out of the check to make it more concise.
```javascript
markersWraps.sort((a, b) => {
  var indexA = Number(a.getAttribute('data:realIndex'))
  var indexB = Number(b.getAttribute('data:realIndex'))
  return indexB < indexA
       ? 1
       : indexB > indexA
       ? -1
       : 0
})
```
If we do not do this, using the short style would also improve legibility:
```javascript
markersWraps.sort((a, b) =>
    Number(b.getAttribute('data:realIndex')) < Number(a.getAttribute('data:realIndex'))
  ? 1
  : Number(b.getAttribute('data:realIndex')) > Number(a.getAttribute('data:realIndex'))
  ? -1
  : 0
)
```

junedchhipa (approved):

PR #3093, opened 2022-04-12
Feature/bg gradient

Atif252:
# New Pull Request

Added `fill` support inside the grid area. This allows us to use the same `ApexFill` config options inside the grid to show gradients, images, or patterns. Also added the support to pass an array of strings into `fill.gradient.type` for multiple grid directions

Screenshot 2022-04-12 at 1 13 51 PM

## Type of change

## Checklist:

### DIFF
--- a/samples/react/column/column-gradient-background.html
+++ b/samples/react/column/column-gradient-background.html
@@ -1,0 +1,167 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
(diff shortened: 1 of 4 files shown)


junedchhipa:
Thanks for the contribution, @Atif252 
A chart can have a transparent background and the background of the chart can be controlled with HTML/CSS independently.
I don't see much value in bringing custom background only for `grid`

PR #3231, opened 2022-06-23
fix: fixed autoScaleY for all zero values in array

lucacapocci94-dev:
# New Pull Request

My fix contain the solution at autoScaleYaxis that for all zero values in a series not show line chart

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/src/modules/Scales.js
+++ b/src/modules/Scales.js
@@ -564,8 +564,10 @@ export default class Range {
         min *= min < 0 ? 1.1 : 0.9
         max *= max < 0 ? 0.9 : 1.1
 
-        if (min === 0) min -= -1
-        if (max === 0) max += 1
+        if (min === 0 && max === 0) {
+          min = -1
+          max = 1
+        }
 
         if (max < 0 && max < initialMax) {
           max = initialMax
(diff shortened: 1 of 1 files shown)


junedchhipa:
What problem does this PR solve? Can you create a Codepen illustrating the fix?

junedchhipa:
Instead of checking min === 0 and max === 0 separately, you should check both in a single condition.

lucacapocci94-dev:
The problem it solves is the following: When there are all zero values ​​in a series, I add an arbitrary minimum and maximum so that the line with zero values ​​is visible.  While to date the autoScale, as I have seen in the code, multiplying a minimum and a maximum equal to zero by "0.9" or "1.1", will always be 0 and therefore the line will never be visible.  Also I agree to put a single control because it is even fairer, i.e. only when in a series the minimum and maximum of the values ​​are both equal to zero, an arbitrary value must be set.

lucacapocci94-dev:
https://user-images.githubusercontent.com/82463600/175947191-6586f93f-39ee-4975-89a1-311c66e0096c.mov

With video you understand better

junedchhipa (approved):

PR #3318, opened 2022-08-18
Handle click on label

### LINKED ISSUE #2324: Click on datalabel should give dataPointIndex
I have a column chart similar to sample distributed-columns. 
But as we have large differences in numbers (one column can have value 2 and another value 13000) then the small columns are not visible. So we have a dataLabel on top with and offset. 
We need to have the columns clickable and for usability also the labels. But I don't get the dataPointIndex in the clickevent. I get it for the columns but not the dataLabel.

I would like the dataPointIndex to be included in the event when clicking on dataLabel the same way as it is when clicking on the column.

**Describe alternatives you've considered**	
I have tried different solutions to hook up listners and so but I can't get it to work and I do only have the value of the column, not the serie or the id for the column.
Alternative would be to use some other chart. But apexchart has been top notch otherwise for me.

**Additional context**	
image

### LINKED ISSUE #3038: Clickable annotations
## Summary

Making annotations clickable and knowing which annotation by getting for example its xaxis index and/or timestamp would be very handy for some use cases.

## API Changes

For example introduce an event called annotationClick() which allows to access different properties of that annotation **as well as its y or x axis index or timestamp to be able to identify them**.

## Intended Use Case

There are anomalies marked with annotations. A click would allow to validate them or add some additional text information.

alik0211:
# Handle click on label

I have added a click event for x-axis labels. The event provides information about a `labelIndex` (in most cases, it corresponds to the ``dataPointIndex`` from event `click`)

image

## Type of change

## Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,16 +20,15 @@ npm run dev
 
 You will now have a fully functioning local build of this library ready to be used. **Leave the `start` script running**, and continue with a new Terminal/shell window.
 
-
 ## Send your changes back to us! :revolving_hearts:
 
 We'd love for you to contribute your changes back to `apexcharts.js`! To do that, it would be great if you could commit your changes to a separate feature branch and open a Pull Request for those changes.
 
-Point your feature branch to use the `master` branch as the base of this PR. The exact commands used depends on how you've setup your local git copy, but the flow could look like this:
+Point your feature branch to use the `main` branch as the base of this PR. The exact commands used depends on how you've setup your local git copy, but the flow could look like this:
 
 ```sh
 # Inside your own copy of `apexcharts.js` package...
-git checkout --branch feature/branch-name-here upstream/master
+git checkout --branch feature/branch-name-here upstream/main
 # Then hack away, and commit your changes:
 git add -A
 git commit -m "Few words about the changes I did"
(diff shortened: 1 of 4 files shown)


brianlagunas:
Have you testing this feature with the other chart types?

alik0211:
Yes. This only works for horizontal bars.

brianlagunas:
Thank you for the PR, but we cannot accept a feature that only works for one chart type. If you would like to add support for all other chart types, please update this PR. Otherwise I will have to close it. Let me know what you would like to do.

alik0211:
I have added support for other chart types. The PR has been updated.

junedchhipa:
`labelClick` is a vague name that can mean any kind of label click. We have x-axis labels, data labels, annotations, etc.
Please rename the property to `yAxisLabelClick` to make it more meaningful.

alik0211:
Good point, thank you! Renamed to `xAxisLabelClick` because it is an event for the x-axis labels.

trylaarsdam:
Does this actually fix #3038? It seems like this PR only provides a click handler for x-axis labels, not annotations. Maybe I missed something though.

alik0211:
I'm sorry, apparently I got the annotations and labels mixed up

alik0211:
@junedchhipa what else do I need to get approved?

alik0211:
I have made as few changes as possible to add this functionality.

junedchhipa (approved):

PR #3480, opened 2022-11-10
Add top position for tooltip

### LINKED ISSUE #3257: Set the position of tooltips
Currently the tooltip's position is determined automatically and can't be controlled through configuration. 
For example in the following screenshot the tooltip is positioned to the top-left of the marker.
image

If I want the tooltip to be centered above the marker like in the screenshot below this is not possible.
image

Since the tooltip is rendered as html it would have been possible at least to change the position with css, but since the automatic positioning is inconsistent (sometimes to the left of the marker, sometimes to the right etc) It's not possible to give the element css that will handle all cases.

I suggest adding the ability to set the tooltip's position as centered relative to the marker in a selected direction (top, bottom, left, right)

Grafikart:
# New Pull Request

Add a new option "position" for tooltip to place the tooltip on top of the point instead of the side. I mirrored `fixed` for the options : 

```js
{
  // ...
  tooltip: {
    absolute: {
      position: 'top',
      offsetY: -13,
    }
  }
}
```

Fixes #3257

## Type of change

## Checklist:

### DIFF
--- a/src/modules/tooltip/Position.js
+++ b/src/modules/tooltip/Position.js
@@ -176,6 +176,16 @@ export default class Position {
 
     let x = parseFloat(cx) + pointR + 5
     let y = parseFloat(cy) + pointR / 2 // - tooltipRect.ttHeight / 2
+    let offsetX = 0
+    let offsetY = 0
+
+    if (w.config.tooltip.absolute && w.config.tooltip.absolute.offsetX) {
+      offsetX = w.config.tooltip.absolute.offsetX || 0
+    }
+
+    if (w.config.tooltip.absolute && w.config.tooltip.absolute.offsetY) {
+      offsetY = w.config.tooltip.absolute.offsetY || 0
+    }
 
     if (x > w.globals.gridWidth / 2) {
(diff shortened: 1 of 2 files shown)


junedchhipa:
Hey @Grafikart 
Thank you for your contribution. Can you just remove the `dist` files as those files only need to be generated when releasing a new version?
Thanks!

Grafikart:
@junedchhipa I removed the dist files

junedchhipa:
@Grafikart  Thanks for removing the `dist` files. As I tested your code, it only works for line/area/scatter charts.
Can you also make this change for column/horizontal bar (and any other relevant) charts as well to make it consistent?

Grafikart:
@junedchhipa Ok i'll find time to work on that, is there a test page in the repository to try these case or do I have to setup a project on the side ?

junedchhipa:
You can use the [samples](https://github.com/apexcharts/apexcharts.js/tree/main/samples) directory which uses the development build of the library to instantly test your changes.

Issue #3646, opened 2023-02-09
Annotation Position


canblt:
### Description

By setting the annotation Postion to 'back', the position will set to front

### Steps to Reproduce

1. 
2. 
3. 

### Reproduction Link

I havnt one its to specail
https://codesandbox.io/s/6yoqonyo7r

canblt:
I have updated The issue Information and added a Reproduction Link

sacalata:
Possibly related to #3654 https://github.com/apexcharts/apexcharts.js/issues/3654

mygpse:
Hey, change the position of the condition in the apexcharts.js line from 32183 to 32137

`if (w.config.annotations.position === 'back') {
     w.globals.dom.Paper.add(w.globals.dom.elAnnotations);
     me.annotations.drawAxesAnnotations();
}`

canblt:
what do you mean with 32183 to 32137

mygpse:
Move the code snippet
`if (w.config.annotations.position === 'back') { 
w.globals.dom.Paper.add(w.globals.dom.elAnnotations); 
me.annotations.drawAxesAnnotations(); 
}`
from line 32183 to line 32137. 

position: 'back' will work

Issue #3654, opened 2023-02-13
3.37.0 Broke annotations on category x-axis 


brodrigz:
### Description
Annotations on an category x-axis are no longer rendered on charts in the latest 3.37.0 release.

This can be observed in the [documentation ](https://apexcharts.com/javascript-chart-demos/column-charts/column-with-rotated-xaxis-labels/)

### Steps to Reproduce

1. Create a chart with the following code on version 3.37.0 :
```
var options = {
          series: [{
          name: 'Servings',
          data: [44, 55, 41, 67, 22, 43, 21, 33, 45, 31, 87, 65, 35]
        }],
          annotations: {
          points: [{
            x: 'Bananas',
            seriesIndex: 0,
            label: {
              borderColor: '#775DD0',
              offsetY: 0,
              style: {
                color: '#fff',
                background: '#775DD0',
              },
              text: 'Bananas are good',
            }
          }]
        },
        chart: {
          height: 350,
          type: 'bar',
        },
        plotOptions: {
          bar: {
            borderRadius: 10,
            columnWidth: '50%',
          }
        },
        dataLabels: {
          enabled: false
        },
        stroke: {
          width: 2
        },
        
        grid: {
          row: {
            colors: ['#fff', '#f2f2f2']
          }
        },
        xaxis: {
          labels: {
            rotate: -45
          },
          categories: ['Apples', 'Oranges', 'Strawberries', 'Pineapples', 'Mangoes', 'Bananas',
            'Blackberries', 'Pears', 'Watermelons', 'Cherries', 'Pomegranates', 'Tangerines', 'Papayas'
          ],
          tickPlacement: 'on'
        },
        yaxis: {
          title: {
            text: 'Servings',
          },
        },
        fill: {
          type: 'gradient',
          gradient: {
            shade: 'light',
            type: "horizontal",
            shadeIntensity: 0.25,
            gradientToColors: undefined,
            inverseColors: true,
            opacityFrom: 0.85,
            opacityTo: 0.85,
            stops: [50, 0, 100]
          },
        }
        };

        var chart = new ApexCharts(document.querySelector("#chart"), options);
        chart.render();
```
      

### Expected Behavior
Annotations are rendered on the chart
### Actual Behavior
Annotations are NOT rendered on the chart

#### PREVIOUS VERSION:
![image](https://user-images.githubusercontent.com/57366583/218559419-2c277c62-4252-4471-99cd-908dae4a0348.png)

#### 3.37.0:
[...]

PR #3705, opened 2023-03-14
Annotation Position back fix 

mygpse:
Annotation Position back fix 
https://github.com/apexcharts/apexcharts.js/issues/3646
https://github.com/apexcharts/apexcharts.js/issues/3654


junedchhipa:
annotation positioning is soon going to be deprecated as it causes other issues with positioning.
Nonetheless, thanks for your contribution.

PR #4143, opened 2023-12-07
Progress: Apply a border radius to the first and last elements of the stacked columns/bars

### LINKED ISSUE #4113: Apply a border radius to the first and last elements of the stacked columns/bars.
## Summary
There should be a way to round the first and last elements of stacked columns/bars. Currently, I can only apply border radius to either the last element or all elements in the stack. I aim to round only the first and last elements.

## API Changes
As per the documentation, the available options for the borderRadiusWhenStacked property are currently last and all. I propose adding an additional option called both or another suitable name. This new option would enable us to apply border radius to the first and last elements of the bar/column chart.
```
borderRadiusWhenStacked: "both",
```
With this change, we gain access to both ends of the chart. However, the radius application needs modification to set border-top radius for the last element and border-bottom radius for the first element. To achieve this, borderRadiusApplication needs to accept a single string with two values or an array with two string values.
Example:
```
chart: {
  type: "bar",
  stacked: true,
  plotOptions: {
      bar: {
        borderRadius: 15,
        borderRadiusApplication: "end start", // or  ["end, "start"]
        borderRadiusWhenStacked: "both",
      },
    }
  }
  }
```

## Intended Use Case
This enhancement can be utilized to create a "pill"-shaped stacked bar/column chart.

image

AvaGivone:
# New Pull Request
Overview: Progress made in implementation of borderRadiusWhenStacked: 'both' feature request. Property implemented and correctly accesses only the first and last stack in stacked bar graph. Created samples for dev. Requesting assistance for front-end graphics issue when all series are positive data values 

Backend: 
- Implemented the 'both' parameter for the borderRadiusWhenStacked property. Updated in the same way that the 'all' and 'last' property were implemented.
- Further Changes: None 

Front-End Graphics: 
- Correctly applies the radius to the ends of ONLY the first and last element when 'both' for borderRadiusWhenStacked
- Created two new ApexCharts stacked bar charts to test the both feature (pill-bar.html and both-bar.html) NOTE: pill-bar.html has a negative series whereas both-bar.html they are all positive
- Further Changes: Yes, issues with applying radius to correct side of the first series when all data positive (see below) 

Testing and Samples: 
- Added pill-bar.html to samples/tests screen shots 
- Added both-bar.html to samples for other contributors to see Front-End issue 
- Added screenshots to test files

Next Steps: 
- Find a way to apply the radius to the other end of the first bar (Already tried modifying the SVG path via modifying code logic in Graphics.js, Helpers.js, and Bar.js. Also tried rotating the first series 180 degrees, but changes were consistently applied to the whole stack as opposed to each individual elements in first stack so misaligned/bars did not stagger correctly (See images) 

Screenshots: 
'both' property implemented: 
Screenshot 2023-12-07 at 4 00 31 PM
https://github.com/apexcharts/apexcharts.js/assets/94867851/2bb9e982-94e9-48d4-8ced-cb4d4cc084bd">

Radius Applied to Wrong Side of the first stack when borderRadiusApplication: 'end'

Both Implemented

Using transformations applies to the entire stack as one element as opposed to each individual bar in series: 
Rotation Dilemma

Fixes #4113 

## Type of change
Feature-request 

## Checklist:


junedchhipa:
Thanks for the contribution.
I think, instead of relying on the index for applying the radius, there should be an option in the `series` itself where a serie can have `applyBorderRadius`. This way, we don't have to worry whether the negative series comes first or in between. I will re-visit this later and modify the conditions.

PR #4159, opened 2023-12-20
Fix/4111 skipped second tick xaxis

### LINKED ISSUE #4111: Wrong hours on x-axis labels if dataset starts at round hours
### Description

When the data starts at a round hour, the hours on the datetime x-axis are shifted one hour except for the first label. The labels don't show the correct time, but the tooltip does.

### Reproduction Link

https://codepen.io/Danielvdd1/pen/poGaOqz
The codepen contains a few datasets to see the different results.

### Expected Behavior
The labels on the x-axis should match with the data.
Example: 9, 10, 11, 12, 13, 14

### Actual Behavior
The first label is correct, but the following labels are shifted as if the second label is ignored.
Example: 9, 11, 12, 13, 14, 15
(The tooltip time is correct. It does not match with the x-axis labels.)

### Screenshots
Screenshot

### Notes
This problem does only occur with rounded hours.
A temporary solution is to subtract 1 millisecond from the first datapoint. (Note: Adding 1 millisecond to the first datapoint will still result in the wrong x-axis labels.)
It can be a problem related to rounding the hours in the labels.

Changing datetimeUTC does not fix the problem. It changes the hours everywhere, on the labels as well as the tooltip. The labels after the first label are still wrong.

Avladd:
# New Pull Request

Fixes x-axis generating an incorrect hourly scale when using a full starting hour. 

This [previous merge](https://github.com/apexcharts/apexcharts.js/pull/3998) introduced this bug because the incrementation for the `hour` was changed outside the if statement but not inside.

Fixes #4111 and [this](https://github.com/apexcharts/Blazor-ApexCharts/issues/342)

## Type of change

## Checklist:

### DIFF
--- a/src/modules/TimeScale.js
+++ b/src/modules/TimeScale.js
@@ -625,7 +625,7 @@ class TimeScale {
     if (remainingMins === 60) {
       firstTickPosition = 0
       firstTickValue = firstVal.minHour
-      hour = firstTickValue + 1
+      hour = firstTickValue
     }
 
     let date = currentDate
--- a/tests/unit/timescale.spec.js
+++ b/tests/unit/timescale.spec.js
@@ -417,6 +417,58 @@ describe('Generate TimeScale', () => {
     }
   })
 
+  it('should generate an hour timescale without skipping ticks when start with full hour', () => {
+    const chart = createChart('line', [
+      {
+        data: [0, 1],
+      },
+    ])
+    const timeScale = new TimeScale(chart)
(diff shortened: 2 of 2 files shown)


Avladd:
For reviewers: I have intentionally not squashed commits for you to more easily test the existing bug. Run the unit tests added in the first commit to see them failing, and then compare to solution.

junedchhipa (approved):

Issue #868, opened 2019-08-27
Support for more marker shapes


maasha:
Can we have additional marker shapes besides circles and squares, such as triangles, asterisks, + and x?

And for circles, squares, triangles, is it possible (perhaps already now for the circles and squares) to have open and close version (with path stroke set and optional stroke fill)?

Also, the orientation of markers, they could be rotated a number of degrees to get more different marker shapes

pasztora:
Hi guys,
Just wanted to check in if there has been any progress on this feature-request so far? Would really love to see more markers than 2 to be able to make proper data comparison. Also it would be great if the selected markers would be visible for the legend as well (at the moment it is only a colored dot). Thanks a lot!

maasha:
Look at the number of upvotes on this issue. The `no-issue-activity` tag added by the bot resulting in closing of this issue is a tad unfair!

julienlaglasseadhzm:
anyone has solution to add more marker shape since 2019 ?

flowzen1337:
can this somehow been relooked at? I would really love to see more shapes, i guess its a low hanging fruit for the devs to get more shapes in it. Would be so so nice! Or something like adding font-awesome as possible shape/icon?!

Khalifa1997:
I agree, there should be some more available shapes

tamerwaw2:
we need markers take more shapes, or just a custom image

KrishnaKantwala:
Yes can someone provide if they figure out the solution for the marker shaper. We need more shapes to support inside the apex-charts

junedchhipa:
This has been supported since **v3.51.0**. Ten shape names are accepted by `markers.shape`:

```
circle (default) | square | rect | triangle | diamond | star | sparkle | cross | plus | line
```

`rect` is an alias of `square`, so that is nine distinct shapes. It takes an array too, so each series can have its own:

```js
markers: {
  size: 8,
  shape: ['circle', 'triangle', 'diamond', 'star', 'cross'],
}
```

Per-point overrides work through `markers.discrete[].shape`, and the same set is honoured by the canvas renderer, so shapes survive `chart.renderer: 'canvas'`.

Re-tested on **6.8.0** before posting, one series per shape: all ten render, with ten distinct path geometries and the `shape` attribute set on each marker element.

The commit that added these closed #4569 and did not reference this issue, which is why this thread was never updated and then went stale. Leaving it closed, now with the release recorded. If you need a shape that is not in that list, please open a fresh issue naming it.

PR #4224, opened 2024-02-09
Added 'X' and '+' marker shapes 

MiguelsPizza:
# Added an 'X' and '+' marker shapes 

Fixes # (issue)
[issue 868](https://github.com/apexcharts/apexcharts.js/issues/868)

This was never actually addressed

### DIFF
--- a/src/modules/Graphics.js
+++ b/src/modules/Graphics.js
@@ -731,13 +731,103 @@ class Graphics {
     return elText
   }
 
+  /**
+   * Creates a group with given attributes.
+   * @param {number} x - The x-coordinate of the group.
+   * @param {number} y - The y-coordinate of the group.
+   * @param {Array} lines - The lines to be added to the group.
+   * @param {Object} opts - The options for the group.
+   * @returns {Object} The created group.
+   */
+  createGroupWithAttributes(x, y, lines, opts) {
+    let elPoint = this.group()
+    lines.forEach((line) => elPoint.add(line))
+    elPoint.attr({
+      class: opts.class ? opts.class : '',
+      cy: y,
+      cx: x,
+    })
+    return elPoint
+  }
+
+  /**
+   * Draws a plus sign at the given coordinates.
+   * @param {number} x - The x-coordinate of the plus sign.
+   * @param {number} y - The y-coordinate of the plus sign.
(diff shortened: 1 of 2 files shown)


MiguelsPizza:
> May I suggest to also include "+" sign along with this? And don't include the `dist` folder as it is only generated when a new version is released.

added the plus sign and some JSDoc types

src/modules/Graphics.js
```diff
@@ -731,13 +731,104 @@ class Graphics {
...
+      opts.pointStrokeLineCap
+    )
+
+    return this.createGroupWithAttributes(x, y, [line1, line2], opts)
+  }
+
   drawMarker(x, y, opts) {
     x = x || 0
     let size = opts.pSize || 0
+    console.info({ opts })
```
junedchhipa, line 824: please remove this console

MiguelsPizza: should be good now

junedchhipa (approved):

djohnson-aperture:
I know this has been merged already, but I wonder if + and X markers should use the `pointFillColor` rather than `pointStrokeColor` as done in this change. 
Here's a roughed out example of how the markers differ in their presentation: 
https://codesandbox.io/p/sandbox/cocky-curie-2km9xj?file=%2Fsrc%2FLineChart%2Fdata.ts
I thought I had a way to work around it by using an array of function for the `marker.stroke.colors`, similar to `stroke.colors` but it only handles a string array. I think what I'm asking would have to be done in the code, since there's no way to access the global colors within this context.

djohnson-aperture:
Also, once I started adding stroke color, another issue arose. It seems if you use one of the new markers on a series that does not fill the whole range, you get extra markers drawn at the top of the chart. `showNullDataPoints: false` has no effect on this defect.
It seems that the legacy markers (circle/square) get sized to 0 if their y value is , so they are still "drawn" at the same positions (top left or top right,) but they aren't visible due to the size of 0. The + and `x` aren't mimicking this behavior, and are visible.
https://codesandbox.io/p/sandbox/invisible-markers-line-chart-forked-w4mcym
Interestingly, the markers are positioned at the top above their last value in the series, but only if a single value is missing. If more than one is missing, the markers are placed at the top left.

PR #4227, opened 2024-02-10
Nice Scale refactoring, brush charts refactoring, and several root-cause bug fixes

rosco54:
# New Pull Request

##### `niceScale()`:

The generation of Y axis ticks and labels would sometimes miss the top-most tick, due to inexact JS float precision and a loop exit condition that tested exact equality.

This was the root-cause for several reported issues that I have tested (referenced in the comments alongside code this is now removed):<br/><br/>Issues: 426, 430, 492, 614, 968.

Mostly, the removed or modified code applied small range adjustments that are now no longer required and were masking other root-cause problems, such as (2) below.

##### `getMinYMaxY()`:
-   referenced `chart.type` instead of `series[i].type`, causing it to return incorrect min/max ranges.<br/><br/>
After fix (1), the chart sample: `tests/e2e/snapshots/candlestick/candlestick-line.png` then exposed this bug.

-   erroneous logic in `if` conditionals (incorrect application of deMorgan's First Rule). Effectively this resulted in matching *any* chart types rather than only the targeted types.

-   incorrect or excessive determination of max and not properly determining min at all, particularly for candlestick series, resulting in clipping below the X axis. This was another bug masked by the range adjustments mentioned previously, and now removed.

### Refactor niceScale() - strengthen the mandate of user defined options - other enhancements

`niceScale(yMin, yMax, index = 0)`

*   Motivation:<br/><br/>
To achieve a predictable and reliable interaction between user defined options and the "niceness" rules applied to charts. To make "nice" the default in all cases while still asserting user defined settings.

*   Refactoring objectives:

    1.  options defined by the user must be honoured, with the exceptions only where users choose settings that conflict with themselves. Users must be able to see the options that they set in the resulting charts. E.g. `min: 0, max: 10, stepSize: 2` should all be honoured.<br/><br/>
    Where the user has introduced an unresolvable conflict, the resolution should be not be arbitrary. E.g. `min: 0, max: 10, stepSize: 3` could result in any of the options being adjusted, so a hierarchy of relative importance is used that results in min and max being retained and stepSize adjusted.<br/><br/>
    This has advantages. E.g. a mandatable tickAmount enables the ticks on each Y axis in a multi-axis chart to be aligned with the drawn grid lines, while being scaled dynamically. See `e2e/snapshots/line/brush-charts.png`.

    2.  generated "nice" values should reflect common real-world choices. Uncommon step sizes should be avoided.

* Args reworked for `Scales.niceScale()`. The state previously passed-in is available from the `w.globals` and `w.config` structures and are now retrieved from within the function.

*   Ugly step sizes are avoided, e.g. [3, 4, 6, 7, 8, 9]*10^n.<br/><br/>
If a specific stepSize is required, e.g. 3 or 0.7, it can set explicitly. If the user is unable to predetermine the order of magnitude of the data values, they can set `forceNiceScale: true`, which will cause the user defined stepSize to be normalized to a magnitude that fits. E.g. and a data or min..max range of 5000, and a user defined `stepSize: 2.5` will be normalized to either `25`, `250` or `2500` depending on the number of ticks (tickAmount or default). Conflict resolution will still apply to the normalized user defined value.

*   Scale all Y axes so that their ticks in multi-axis charts align to the grid lines.
*   The number of ticks displayed will be reduced if the svg dimensions become small enough to cause overcrowding. The reduced set of ticks will always be a subset of the same full set.

### New supporting utility functions

*   `Utils.getGCD(a, b, p = 7)` :- return the Greatest Common Divisor of `a` and `b`. Arg `p` is an optional power factor for precision.

*   `Utils.getPrimeFactors(n)` :- `n` must be an integer, return an array containing all prime factors of `n`, in order from lowest to highest.

*   `Utils.mod(a, b, p = 7)` :- return `a` modulo `b`.  Arg `p` is an optional power factor for precision. Required because the JS inbuilt `%` operator doesn't deal well with precision.

### Integrate the above with existing experimental Brush Charts

It may now be possible to remove the alpha status from brush charts.

Make use of existing method `Range.getMinYMaxY()` to determine Y range based on the brush X range. This method already understands how to extract min and max from the different chart types. Introduce `Scales.niceScale()` to scale the target Y axis based on the brush X range.

The experimental `Scales.autoScaleY()` function plus the calling and post call assignments in `Core.setupBrushHandler` are no longer needed.

Target charts with multiple series and/or multiple Y axes are working. Collapsed series are handled normally.

Notes:

1.  The class defined in `Scales.js` has been renamed from "Range" to "Scales" to disambiguate it from the `Range.js` "Range" class.

2.  The method `Scales.autoScaleY()` has been removed.

3.  Bug fix: `Core.setupBrushHandler()` was cloning `w.config.yaxis`instead of `targetChart.w.config.yaxis`. The selection event handler has been reduced to only record the selection X range limits.

### A new sample demonstrating duplicate label removal/hiding

AFAICS duplicate labels can only be properly handled from within a formatter.

-   `tests\source\mixed\duplicate-labels.xml`
Demonstrates how to hide duplicate labels, or any labels that do not accurately represent a tick grid line.
    
## Type of change

- [X] Bug fix (non-breaking change which fixes an issue). Root cause of several previously fixed bugs.
- [X] New feature (non-breaking change which adds functionality). As should be expected, several existing charts display differently BUT  CORRECTLY. Other than the brush charts, which is a work-in-progress and which does not yet call `niceScale()` yet, are visually untouched by these changes. (The zoom redraw immediately obliterates any nice scaling that has been done.)

## Checklist:

Tests are very much expected to fail due to the nature of the changes.
- [] New and existing unit tests pass locally with my changes.

### e2e tests

Mostly subtle differences in charts.

All "failures" are expected and reflect the changes made as described.

Several samples have been minimally modified as appropriate.

### unit tests

See the changes to individual test js files.

All unit tests pass after appropriate revisions to the `expect` results.


junedchhipa:
@rosco54 This is a great contribution 👏🏻 
May I ask not to include the `dist` directory in this PR as these files are only generated when a new version released.

I have gone through the PR once and will probably re-visit to see all the changes.

junedchhipa (approved):

junedchhipa:
@rosco54 The zooming functionality has stopped working after the merge.
Please check this sample - samples/vanilla-js/line/zoomable-timeseries.html

Issue #4237, opened 2024-02-14
The possibility to have an array of series in seriesName of y-axis


MarieNuellas:
## Summary
It would be cool if we could select multiple series in seriesName so in a multiple y-axis chart we can target the scale of a y-axis from multiple series 

## API Changes

Include a list of all API changes, additions, subtractions as would be required by your proposal. These APIs should be considered placeholders, so the naming is not as important as getting the concepts correct. If possible you should include some "example" code of usage of your new API.
With my proposal, y-axis options would pass from 
`seriesName: String`
to
`seriesNames: String | String[]`
or simply
`seriesNames: String[]`

## Intended Use Case
With my proposal, it would permit to add a scale for multiple groups of series, in the exemple there are 2 groups of series, original and compared. it would allow me to make a scale for the two groups
<img width="775" alt="Screenshot 2024-02-14 at 11 27 46" src="https://github.com/apexcharts/apexcharts.js/assets/145000510/8a500755-8c7e-43e7-9ff7-962ba0c8ca6c">

rosco54:
@junedchhipa this request can be closed, unless you're waiting to document it.

PR #4247, opened 2024-02-18
Implement feature request #4237 plus refactor similar scale multi-axis functions

rosco54:
# New feature

Implement feature request #4237 
See new: sample/src/area/multi-axis-with-seriesname-arrays.xml

Note: it is left to the creativity of the user to visually disambiguate which plot is associated with which Y axis.

# Extended application of this feature

The current style of mapping multiple series to a y axis is to include one yaxis config per series but set each yaxis seriesName to the same series name and then set show: false on all axes that are not required.

While not the purpose of #4237, the feature offers a more intuitive alternative that explicitly configures yaxis elements with the series that will be referenced to them (seriesName: []). This only requires including the yaxis elements that will be seen on the chart. This is how it works to the best of my limited knowledge.

Old way:
  yax:	ser
  0:	0
  1:	1
  2:	1
  3:	1
  4:	1

Axes 0..4 are all scaled and all are be rendered unless the axes are show: false. If the chart is stacked, it's assumed that series 1..4 corresponding to the yaxis indices 1..4 are the series' that will be stacked.

## Implementation of the extended feature

Use the new feature yaxis: seriesName: [] to simplify the specification of series inclusion in stacked charts, especially when combined with other series. Essentially, refactor Scales.sameScaleInMultipleAxes() and Scales.setMultipleYScales() to use yaxis: seriesName: [].

New way:
  yax:	ser
  0:	[0]
  1:	[1,2,3,4]

See stacked-column-with-line.xml
and c.f. stacked-column-with-line-new.xml

The new facility is backward compatible with the existing style via code that converts the old data structure to the new one.

If the chart is stacked (a global chart setting), it is assumed that any axis with multiple series is stacked. Multiple yaxes referencing multiple series in a stacked chart is not implemented (or rather doesn't do anything particularly useful yet), however it was noted that it does not break the chart. It essentially does what a user might expect it to do, which is add an additional Y axis.

Fix stacked chart y axis scaling: min and max were computed as the sum of the series min and max, not the min and max of the one-for-one summed datapoints in each series.

## Minor fixes

Additional checks for undefined elements in CoreUtils.extendArrayProps().

Remove the duplicate outlier in boxplot-scatter.xml.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## Checklist:


rosco54:
Please hold off reviewing. There is issue that I need to address.

rosco54:
The issue (an off-by-one mistake) has now been fixed, plus added some unit tests. Should be good to review. Thanks.

junedchhipa (approved):

PR #4327, opened 2024-03-13
If all series types match the chart type it isn't a combo chart

### LINKED ISSUE #4132: Stacked Area Charts Don't Stack
### Description
Prior to 3.44.1, stacked area charts would stack, even with a combo chart. I have a combo chart that has used stacked bars, stacked areas and a single line for over a year. But as of 3.44.1, the bars stack, but the areas do not stack. Even after removing all other series except for the areas and they still won't stack. Even Apexchart's website's demo for stacked areas doesn't stack now.

### Steps to Reproduce

1. Create a simple stacked area chart
2. Observe that chart does not stack (areas overlap instead of stacking)
3. Or review the apexcharts stacked area page (https://apexcharts.com/javascript-chart-demos/area-charts/stacked/) and observe that the areas are not stacked

### Expected Behavior
In the codepen/reproduction link, I would expect the values to stack and plot on the Y value at the sum of all values. For example, the last values in 1999 are 125 and 100, which, stacked, should plot at 225.

In the screenshot of the apex charts website, the values are 20, 11, 10, which stacked would plot at 41.

### Actual Behavior
"Stacked" areas (in codepen and apex charts website screenshot) are overlapping, not stacking.

### Screenshots
This is from the apex charts website to show even the demo is not functioning properly
apex_charts_website_stacked_area_demo

### Reproduction Link
https://codepen.io/krcoats/pen/OJdrLKW

Thank you for the help.

rosco54:
# New Pull Request

The default type for chart is "line" and the default for series is the chart type, therefore, if the types of all series match the chart type, this should not be considered a combo chart. Previously, if any series had a defined type, the chart was considered a combo chart.

Fixes #4132

The CodePen example in the original  bug report will stack correctly with this fix, but as far as I know it doesn't restore the behaviour the reporter claims was working in revisions prior to 3.44.1.

### Before
2fz93k98

### After

t51xbkn1 (1)

## Type of change

## Checklist:

### DIFF
--- a/src/modules/Core.js
+++ b/src/modules/Core.js
@@ -178,9 +178,16 @@ export default class Core {
       i: [],
     }
 
+    let chartType = cnf.chart.type !== undefined ? cnf.chart.type : 'line'
+    // Check if the user has specified a type for any series.
+    let comboCount = 0
     gl.series.map((serie, st) => {
-      let comboCount = 0
-      // if user has specified a particular type for particular series
+      // The default type for chart is "line" and the default for series is the
+      // chart type, therefore, if the types of all series match the chart type,
+      // this should not be considered a combo chart.
+      // Combo charts are explicitly excluded from stacking with the exception
+      // that series of type "bar" can be stacked if the user sets "stackOnlyBar"
+      // true.
       if (typeof ser[st].type !== 'undefined') {
         if (ser[st].type === 'column' || ser[st].type === 'bar') {
           if (gl.series.length > 1 && cnf.plotOptions.bar.horizontal) {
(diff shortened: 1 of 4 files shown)


src/modules/CoreUtils.js
```diff
@@ -8,12 +8,19 @@ class CoreUtils {
...
 
-  static checkComboSeries(series) {
+  static checkComboSeries(series, chartType) {
     let comboCharts = false
     let comboBarCount = 0
     let comboCount = 0
-
-    // if user specified a type in series too, turn on comboCharts flag
+    
+    if (chartType === 'undefined') {
```
junedchhipa, line 16: undefined as a string should be checked as `typeof chartType === 'undefined'`

src/modules/Core.js
```diff
@@ -178,9 +178,16 @@ export default class Core {
       i: [],
     }
 
+    let chartType = cnf.chart.type !== 'undefined' ? cnf.chart.type : 'line'
```
junedchhipa, line 181: undefined as a string should be checked as `typeof cnf.chart.type !== 'undefined'`

junedchhipa (changes requested):

src/modules/CoreUtils.js
rosco54: Thanks. Done

junedchhipa (approved):

PR #4640, opened 2024-08-15
Changes to accommodate issue 4632

### LINKED ISSUE #4632: Inconsistent/buggy yaxis ticks (too many or none)
### Description

yaxis ticks have had some significant changes in version 3.46 and 3.47, and the behavior of the newer versions is worse IMO.

In version 3.45, for my data sets and chart size of 250px height, I was consistently seeing about 5 yaxis ticks.
In version 3.46, with the same data and height, I was suddenly seeing double the ticks. 10 ticks is too many for a 250px height chart, and looks significantly worse IMO.
In version 3.47+, I'm still seeing about 10 ticks, but have noticed that with some data sets, there are zero ticks when there definitely should be some.

I know I can set the number of number of yaxis ticks with the `tickAmount` option, but it would be nice if the default behavior was sufficient. Also, for some strange reason, starting in 3.47.0, if I set the tickAmount, it can result in the first yaxis tick being negative, even though all the values are positive. Again, I'd like the min and max to be automatic/dynamic, not hardcoded to zero.

This could be 3 separate issues, but I think they could all be related to the same function(s).

### Steps to Reproduce

1. Create a few simple line charts with any data you choose. My series values are typically between 5 and 10,000.
2. Compare the yaxis behavior between apexcharts versions 3.45.2, 3.46.0, 3.47.0, and latest.

### Expected Behavior

The number and spacing of yaxis ticks should not clutter the chart, and should be somewhat consistent over time.

### Actual Behavior

The number and spacing of yaxis ticks has changed significantly in versions 3.46 and 3.47, and the ticks have cluttered the chart in these later versions. Sometimes the ticks are missing altogether.

### Screenshots

This matrix demonstrates how things have changed over time, and the bug in 3.47+ where there are no ticks:

Apex Chart Version Comparisons

### Reproduction Link

Chart with missing ticks: https://codepen.io/jsmartt/pen/OJeOLwr

Comparisons between version 3.45.2 and latest with the same data: https://codepen.io/jsmartt/pen/poXdzxJ & https://codepen.io/jsmartt/pen/wvLPvrK

Chart with yaxis min set to negative number, even though all values are positive: https://codepen.io/jsmartt/pen/mdZqdqZ (broke in 3.47.0)

rosco54:
1. Unconditionally select default initial axis ticks based on SVG size.
2. Unconditionally, if not user specified, snap yMin or yMax to zero if already close.
3. Re-order yaxes in one sample chart to prioritize stacked column visual resolution. Add comment to draw attention to this feature of multi axis charts.
4. In Scales.niceScale: Check that maxTicks is always valid (suspect some Unit tests don't set SVG dimensions).
5. Set tickAmount explicitly in two unit tests to compensate for variation in scaling following the above changes.

Fixes #4632

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/samples/react/column/stacked-column-with-line-new.html
+++ b/samples/react/column/stacked-column-with-line-new.html
@@ -295,13 +295,19 @@
                   }
                 }
               },
+              // In order to align the ticks of multiple yaxes with the charts horizontal grid
+              // lines, the scaling of the first [unhidden] yaxis determines the number
+              // of grid lines and hence the number of ticks for all subsequent yaxes, unless
+              // they specify a tickAmount.
+              // We want the stacked bars to have the best possible resolution
+              // visually so they have been listed first.
               yaxis: [
+                {
+                  seriesName: ['Bar1','Bar2','Bar3','bar4']
+                },
                 {
                   seriesName: 'Line',
                   opposite: true
-                },
-                {
-                  seriesName: ['Bar1','Bar2','Bar3','bar4']
                 }
               ],
               legend: {
--- a/samples/source/column/stacked-column-with-line-new.xml
+++ b/samples/source/column/stacked-column-with-line-new.xml
@@ -76,13 +76,19 @@ xaxis: {
     }
   }
 },
+// In order to align the ticks of multiple yaxes with the charts horizontal grid
+// lines, the scaling of the first [unhidden] yaxis determines the number
+// of grid lines and hence the number of ticks for all subsequent yaxes, unless
+// they specify a tickAmount.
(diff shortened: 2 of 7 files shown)


rosco54:
@junedchhipa There are e2e images diff failures that are expected but some that I don't expect (like the heatmap diffs). I assume you will be looking at these before accepting the PR. Let me know if there's anything that I need to do at this end please.

junedchhipa (approved):

PR #4691, opened 2024-09-09
Added support for the Angular compiler.

jeroen1602:
# New Pull Request

The Angular compiler looks for the main field using a few fields in the package.json. The order it uses is:
- es2020
- es2015
- browser
- module
- main

The first entry the Angular compiler would find before this change was the 'main' entry point. This is however compiled as `common js` and Angular is looking for an `esm` so it would complain (it would still work though). Adding `es2020` as an entry point fixes this problem.

Fixes [ng-apexcharts#92](https://github.com/apexcharts/ng-apexcharts/issues/92)

## Type of change

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "url": "https://github.com/apexcharts/apexcharts.js.git"
   },
   "main": "dist/apexcharts.common.js",
+  "es2020": "dist/apexcharts.esm.js",
   "unpkg": "dist/apexcharts.js",
   "jsdelivr": "dist/apexcharts.js",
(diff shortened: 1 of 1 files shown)


jeroen1602:
Using `module` in the package.json may be more correct and also help other compilers. But I couldn't really find any good documentation on it.

junedchhipa:
We disabled esm builds because we had some issues in the past. The export from Svg.js is being exported instead of ApexCharts. We are in the process of upgrading svg.js and that will potentially fix the export issue. Once it is done, I will revisit this PR

junedchhipa (approved):

PR #4782, opened 2024-10-31
Fix isolated point in line chart not closing path

bautisflow:
With a Line chart representing a time series, isolated points were trying to close a path. This PR fixes this issue by checking that the point is unique using the already present logic that was handling isolated points in between `null` values.

Current version (4.0.0) jsfiddle showcasing the issue: https://jsfiddle.net/kocz16s2/ 

## Type of change

## Checklist:

### DIFF
--- a/src/charts/Line.js
+++ b/src/charts/Line.js
@@ -953,7 +953,7 @@ class Line {
               areaPath = graphics.move(pX, pY)
 
               // Check for single isolated point
-              if (series[i][j + 1] === null) {
+              if (series[i][j + 1] === null || typeof series[i][j + 1] === 'undefined') {
                 linePaths.push(linePath)
                 areaPaths.push(areaPath)
                 // Stay in pathState = 0;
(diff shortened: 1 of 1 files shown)


junedchhipa (approved):

junedchhipa:
This PR caused this bug - https://github.com/apexcharts/apexcharts.js/issues/4855
Hence, I am reverting this and will look for other ways to solve it.

bautisflow:
Hi @junedchhipa 
I think a solution for the linked issue and in order to keep this PR is to check for the length of the series before trying to create a path.

https://github.com/apexcharts/apexcharts.js/blob/b9c58d7a855b414381c844c16b27be122e303e62/src/charts/Line.js#L589

By checking it just as this loop is entered both issues could be fixed, and as far as I got it does not seem that it is required to loop over the maximum range of dataPoints for the series for the hidden one.

Let me now if this solution suits the project and if you want me to create a  PR

Regards

junedchhipa:
Sure, please create a new PR

PR #4909, opened 2025-01-09
range: Fix getting decimal part of numbers close to 0

obiwac:
# New Pull Request

Previously, isFloat was being called on val before noExponents. This meant that if a value was particularly close to zero, it might pass the isFloat check but then be stripped down to just 0 by noExponents, which meant that getting the second element after splitting around the decimal point would fail.

This patch just does noExponent first so that if the value ends up being a round number, it doesn't enter that branch.

## Type of change

## Checklist:

### DIFF
--- a/src/modules/Range.js
+++ b/src/modules/Range.js
@@ -167,8 +167,8 @@ class Range {
           }
           highestY = maxY
 
+          val = Utils.noExponents(val)
           if (Utils.isFloat(val)) {
-            val = Utils.noExponents(val)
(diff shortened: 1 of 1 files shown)


twalk-bnewie (approved):

mjchang (approved):
Thanks - coincidentally just encountered this too https://github.com/apexcharts/apexcharts.js/issues/4911

junedchhipa (approved):

Issue #4897, opened 2025-01-05
v4 crops chart lines and other elements


ivictbor:
### Description

On version 4.3.0 Chart lines are cropped

### Steps to Reproduce

1. Open this CodePan https://codepen.io/Ivan-Borshchov/pen/VYZyZwo
2. Pay attention to line at square edges, they are like cropped by rect area:

![image](https://github.com/user-attachments/assets/4cce7299-550d-48ef-b0d7-0c06a438f31f)

### Expected Behavior

Line should not be cropped. For example 3.52.0 works fine:

![image](https://github.com/user-attachments/assets/427d8af4-8114-4430-9953-a9600ab9903e)

### Actual Behavior

Line edges are cropped:

![image](https://github.com/user-attachments/assets/4cce7299-550d-48ef-b0d7-0c06a438f31f)

### Screenshots

Added above

### Reproduction Link

https://codepen.io/Ivan-Borshchov/pen/VYZyZwo

Please note that issue reproduces even without any of my specific settings, here is another codepan with very minimal settings:

https://codepen.io/Ivan-Borshchov/pen/jENYNPG

Edges are cropped still:

![image](https://github.com/user-attachments/assets/77a521c4-3def-4afd-aa47-e83b4860de63)

And here is same minimal pan on 3.52: https://codepen.io/Ivan-Borshchov/pen/bNbabEq

And it is working as expected:

![image](https://github.com/user-attachments/assets/3e3337e5-5016-451e-9974-31e498db9a1f)

PR #4955, opened 2025-02-18
Fix issue where bars and data labels are drawn beyond the grid in bar chart

### LINKED ISSUE #4956: Issue where bars and data labels are drawn beyond the grid in bar chart
### Description

There is an issue in the bar chart where bars and data labels are drawn even for data points that exceed the currently visible x-axis range.

As a result (**error list**):
**(1)** Datalabels overlap with the y-axis
**(2)** When grid padding is applied, bars and datalabels are drawn beyond the grid into the outer area.
**(3)** The boundaries where bars and data labels are drawn are inconsistent.

### Steps to Reproduce

1. Create a bar chart
2. Set the x-axis min and max values to be narrower than the range of the series' x values.
3.  
   - To observe **error(1)**, set the chart width to be responsive and resize it. The issue occurs depending on the chart's width.  
   - To observe **error(2)**, apply a large grid padding value.  
   - **error(3)** can be observed in both cases above.

### Expected Behavior

Both bars and data labels should be displayed according to the currently visible x-axis range.

### Actual Behavior

Just as explained above.

### Screenshots

Image
Image

### Reproduction Link

https://codepen.io/lihwzlrq-the-styleful/pen/azbvXdE
https://codepen.io/lihwzlrq-the-styleful/pen/PwoPLYz

matchlessNostril:
# New Pull Request

Fixes #4956 

1. The data label issue was resolved by filtering out any data points that fall outside the currently visible x-axis range and setting the data label text to an empty string (`""`) so that labels aren’t drawn. 
image

2. The problem of bars being drawn beyond the grid was fixed by making `elGridRectBar` occupy the x-axis width plus half the bar width on each side. 
image

However, I have some questions about how the code was originally written regarding point (2). I’m unsure about the logic behind the `x1`, `x2`, `y1`, and `y2` parameters passed to the `drawRect` function for both `elGridRect` and `elGridRectBar`.

First, `elGridRect` seems to have been changed from `(0, 0, gl.gridWidth, gl.gridHeight)` to `(-strokeSize / 2 - 2, -strokeSize / 2 - 2, gl.gridWidth + strokeSize + 4, gl.gridHeight + strokeSize + 4)` due to issue #4897. I think `strokeSize` has no relation to width at all, and I’m not sure why `-2` or `+4` were used. Since I confirmed that issue #4897 is still resolved with my update, I changed it accordingly.

As for `elGridRectBar`, the original code used `strokeSize`, but I removed it because I couldn’t see why it was necessary. It also used `barWidthLeft` and `barWidthRight`, and I don’t understand why `grid padding` was included in those calculations. If `elGridRectBar` is extended by that padding, wouldn’t bars appear in the padded area?

Looking at `modules/dimensions/Grid.js`, `barPadForNumericAxis` appears to be the bar width that was computed, but in my case, that value was much smaller than the actual bar width. So I calculated half of the actual bar width and stored it in a variable called `elGridRectBarSide`, which I then used in the `drawRect` call.

If there’s anything important I might have overlooked regarding these points, please let me know. Thank you.

## Type of change

## Checklist:

### DIFF
--- a/src/charts/common/bar/DataLabels.js
+++ b/src/charts/common/bar/DataLabels.js
@@ -56,6 +56,20 @@ export default class BarDataLabels {
       bcy = y + parseFloat(barHeight * visibleSeries)
     }
 
+    const xAxisScale = w.globals.xAxisScale
+    const seriesX = w.globals.seriesX[realIndex]
+    const seriesY =
+      w.config.chart.stackType === '100%'
+        ? series[realIndex]
+        : w.globals.series[realIndex]
+
+    const valY = this.barCtx.isRangeBar
+      ? [y1, y2]
+      : w.config.chart.stackType === '100%'
+      ? series[realIndex][j]
+      : w.globals.series[realIndex][j]
+    const valX = seriesX[seriesY.indexOf(valY)]
+
     let dataLabels = null
     let totalDataLabels = null
     let dataLabelsX = x
@@ -86,11 +100,9 @@ export default class BarDataLabels {
       height: 0,
     }
     if (w.config.dataLabels.enabled) {
-      const yLabel = w.globals.series[i][j]
-
       textRects = graphics.getTextRects(
         w.config.dataLabels.formatter
-          ? w.config.dataLabels.formatter(yLabel, {
+          ? w.config.dataLabels.formatter(valY, {
               ...w,
               seriesIndex: i,
               dataPointIndex: j,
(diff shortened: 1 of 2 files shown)


junedchhipa:
Why does the 2nd screenshot has so much empty space on the right hand side?
Please do a `npm run e2e` and check if all tests pass successfully with your change.
If yes, please let me know, so I will review.

PR #4974, opened 2025-03-07
Fix y-axis formatter retrieval by series name

### LINKED ISSUE #4966: Tooltip formatting not respecting seriesName of yAxis
### Description

If you have custom formatter set on yAxis and use seriesName to target specific series all works fine except when tooltip is shown, which is not using correct formatter.

### Steps to Reproduce

1. Define multiple series
2. Define multiple yAxis with different formatters
3. Use seriesName on yAxis to target correct series
4. Observe that tooltip does not use correct formatter

### Expected Behavior
3rd row in tooltip should end with S2.

### Actual Behavior
3rd row in tooltip ends with S1 instead of S2.

### Screenshots

Image

### Reproduction Link

https://codepen.io/David-Karla/pen/dPyXjKL

hhjeee:
# New Pull Request

Fix #4966 

This PR modifies the way the y-axis formatter is retrieved in getFormatters(i).

Previously, the formatter was obtained from w.globals.yLabelFormatters[i], which relied on the index. This approach caused issues when the yaxis array was shorter than the series array, especially in cases where some yaxis items had seriesName values in an array format (as seen in the issue example).

To resolve this, I introduced a new function, getFormatterBySeriesName(seriesIndex, w), which finds the appropriate formatter by matching the seriesName within the yaxis array. This ensures that the correct formatter is applied based on the series name rather than relying on array indices.

## Type of change

## Checklist:

### DIFF
--- a/src/modules/tooltip/Labels.js
+++ b/src/modules/tooltip/Labels.js
@@ -202,10 +202,30 @@ export default class Labels {
     }
   }
 
+  getFormatterBySeriesName(seriesIndex, w) {
+    const seriesName = w.config.series[seriesIndex].name
+    const yAxis = w.config.yaxis
+
+    for (let i = 0; i < yAxis.length; i++) {
+      const yAxisItem = yAxis[i]
+
+      if (Array.isArray(yAxisItem.seriesName)) {
+        if (yAxisItem.seriesName.includes(seriesName)) {
+          return yAxisItem.labels.formatter
+        }
+      } else if (yAxisItem.seriesName === seriesName) {
+        return yAxisItem.labels.formatter
+      }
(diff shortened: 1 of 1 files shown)


hhjeee:
Hi! Just wanted to follow up on this PR.

I understand that introducing a new function like `getFormatterBySeriesName` might require careful consideration. I’m also totally open to alternative implementations if this approach isn’t the most efficient.

That said, I do believe the current index-based logic for retrieving y-axis formatters (`w.globals.yLabelFormatters[i]`) should be revisited, as it can lead to mismatches when the yaxis array doesn’t align with the series array — especially when seriesName is an array.

PR #5039, opened 2025-05-15
 Remove old chart instance when screen resizes to prevent memory leak

### LINKED ISSUE #5028: Memory Leak When Using `responsive` Option in Charts on Extreme Resize
### Description

When using the `responsive` option in ApexCharts, resizing the browser window repeatedly and aggressively causes a **memory leak**. Charts without the `responsive` config do not have this issue.

In the provided reproduction, only `chart1`, `chart30`, and `chart2` have the `responsive` configuration. Removing these charts (or just removing their `responsive` option) resolves the problem, confirming that the issue lies within how responsive behavior is handled.

### Steps to Reproduce

1. Open the CodePen linked below.
2. Continuously resize the browser window back and forth (small → large → small).
3. Monitor the memory usage in Firefox Developer Tools (Performance or Memory tab).
4. Observe that memory usage increases over time and is not reclaimed, indicating a leak.

### Expected Behavior

Charts should update responsively without leaking memory, even during frequent or extreme window resizes.

### Actual Behavior

Charts with `responsive` configuration cause memory usage to grow continuously with each resize. This can lead to sluggish performance or browser tab crashes over time.

### Screenshots
[https://tinyurl.com/2a2cceuf](https://tinyurl.com/2a2cceuf)

### Reproduction Link
[https://codepen.io/Dharmil-Decode/pen/WbbXyEo](https://codepen.io/Dharmil-Decode/pen/WbbXyEo)

cocarmon:
# New Pull Request

This PR fixes the issue of the memory leak described in #5028 . 

Each time the screen is resized, a new chart is created and added to the global `Apex._chartInstances` array. However, old chart instances were not being cleaned up on window resize, which led to a significant increase in memory usage over multiple resizes.

I extracted the logic used in the `ApexCharts ` class’s `destroy()` method to remove old chart instances, moved it into a private method (`_removeOldChartInstance()`), and now call that method from both `update()` and `destroy()`.

Fixes #5028

Below is a comparison of the memory after resizing the window 7 times. After multiple tests the average is a ~97% reduction in memory usage.

Before:

After:

## Type of change

## Checklist:

### DIFF
--- a/src/apexcharts.js
+++ b/src/apexcharts.js
@@ -399,15 +399,9 @@ export default class ApexCharts {
     window.removeEventListener('resize', this.windowResizeHandler)
 
     removeResizeListener(this.el.parentNode, this.parentResizeHandler)
-    // remove the chart's instance from the global Apex._chartInstances
-    const chartID = this.w.config.chart.id
-    if (chartID) {
-      Apex._chartInstances.forEach((c, i) => {
-        if (c.id === Utils.escapeString(chartID)) {
-          Apex._chartInstances.splice(i, 1)
-        }
-      })
-    }
+
+    this._removeOldChartInstance()
+
     new Destroy(this.ctx).clear({ isUpdating: false })
   }
 
@@ -544,6 +538,8 @@ export default class ApexCharts {
         // Options are identical, skip the update
         return resolve(this)
       }
+
+      this._removeOldChartInstance()
 
       this.lastUpdateOptions = Utils.clone(options)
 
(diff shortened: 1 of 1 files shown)


junedchhipa:
@cocarmon  Can you create a codepen for the issue where _chartInstances is not being cleared during resize?
When I logged _chartInstances during resize, it's always being cleared.

cocarmon:
This fixes the memory leak, but only because it nukes the _chartinstances which isn't ideal as they aren't repopulated on every resize. I'm starting a new job soon so I have had limited amount of time to dedicate to this, I will look more into it though.

PR #5045, opened 2025-05-26
When the numeric range is small, show one tick for every integer, ins…

### LINKED ISSUE #5015: Line chart: Numeric X-axis missing last tick, points misaligned
### Description

The x-axis in line charts doesn't display all tick values correctly, especially when using numeric values as x values across multiple series. The last x value (5) is missing from the axis, and data points are misaligned with their corresponding tick marks.

### Steps to Reproduce

1. Create a line chart with multiple series
2. Use numeric values as x values (1, 2, 4, 5)
3. Observe the x-axis rendering and data point alignment

### Expected Behavior

The x-axis should display all unique x values (1, 2, 4, 5) as tick marks, and data points should align with their corresponding tick marks.

### Actual Behavior

The x-axis only shows 1, 2, 3, 4 as ticks, omitting 5. Additionally:
- The data point for 2 appears before the 2 tick mark
- The data point for 5 is incorrectly positioned above the 4 tick mark

### Screenshots

Image

### Reproduction Link

https://codepen.io/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx-/pen/pvvygbg?editors=0010

```js
var options = {
  "chart": {
    "type": "line"
  },
  "series": [
    {
      "name": "A",
      "data": [
        {
          "x": 1,
          "y": 35
        },
        {
          "x": 2,
          "y": 15
        }
      ],
    },
    {
      "name": "B",
      "data": [
        {
          "x": 4,
          "y": 11
        },
        {
          "x": 5,
          "y": 22
        }
      ],
    }
  ]
};

chart.render();
```

Initially reported in https://github.com/sqlpage/SQLPage/issues/883

lovasoa:
fixes #5015

The code is quite hairy, so I'm not sure this is the proper fix. 

And there may be additional bugs in addition to this one (even with the odd values passed as `xaxisLabels`, it should probably never show a tick with a label that does not match its position).

@junedchhipa

### DIFF
--- a/src/modules/Range.js
+++ b/src/modules/Range.js
@@ -428,7 +428,8 @@ class Range {
         if (gl.isXNumeric) {
           const diff = Math.round(gl.maxX - gl.minX)
           if (diff < 30) {
-            ticks = diff - 1
+            // When numeric range is small, show a tick for every integer
+            ticks = diff
           }
         }
       } else {


lovasoa:
@junedchhipa , what do you think of this ?

junedchhipa (approved):

PR #5200, opened 2026-04-05
fix(annotations): preserve original x value for label lookup in getStringX

cyphercodes:
## Description

When `xaxis.convertedCatToNumeric` is true, the `getStringX` function in the annotations module was modifying the `x` parameter directly, causing the subsequent label lookup to fail. The `labels` array contains the original string values, but after the conversion, `x` became a numeric index (categoryLabels index + 1) which wouldn't match any label in the array.

This fix introduces a separate `lookupValue` variable to preserve the original `x` value while still performing the category label lookup correctly.

## Related Issue

Fixes #5198

## Type of Change

## Testing

- Added unit tests for the `convertedCatToNumeric` scenario
- Verified the fix handles both string category labels and numeric string inputs

## Checklist

- [x] I have added tests that prove my fix is effective

### DIFF
--- a/src/modules/annotations/Helpers.js
+++ b/src/modules/annotations/Helpers.js
@@ -295,12 +295,13 @@ export default class Helpers {
   getStringX(x) {
     const w = this.w
     let rX = x
+    let lookupValue = x
 
     if (
       w.config.xaxis.convertedCatToNumeric &&
       w.labelData.categoryLabels.length
     ) {
-      x = w.labelData.categoryLabels.indexOf(String(x)) + 1
+      lookupValue = w.labelData.categoryLabels.indexOf(String(x)) + 1
     }
 
     const catIndex = w.labelData.labels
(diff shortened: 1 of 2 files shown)


cyphercodes:
Reviewed #5201 against this PR.

#5201 has already been merged into `main` (merge commit `7dd405b`) and current `main` now fixes the #5198 regression in `getStringX` by comparing category labels and `x` via `String(...)` before resolving the numeric category index. It also includes a regression test for numeric x-axis annotations on line charts with simple array data.

Local verification on current `main`:
- `yarn install --frozen-lockfile --ignore-scripts --non-interactive`
- `yarn vitest run tests/unit/xaxis-annotations.spec.js -t "regression #5198" --testTimeout 30000 --hookTimeout 30000` ✅

I did not push an update to this PR branch because the remaining changes conflict with the already-merged #5201 implementation and would mostly duplicate/replace the merged fix/test. If additional `annotation-helpers.spec.js` coverage is still wanted, this PR can be reworked narrowly for tests only.

PR #5212, opened 2026-05-04
fix(responsive): deep-merge yaxis array entries with base config on breakpoint

### LINKED ISSUE #5211: Base yaxis options not merged correctly with responsive yaxis options
### Steps to Reproduce

1. Create a chart with some config specified for the yaxis 
2. Add responsive config with yaxis defined, but different keys than your base config
3. Observe chart above and below specified breakpoint

### Expected Behavior
Y axis config is a merged set of options, with the responsive yaxis config overriding the base yaxis config only for specified keys. Base yaxis config for non-overridden keys still applies.

### Actual Behavior
Y axis only retains settings that are set in the responsive yaxis config. Base yaxis config ignored.

### Reproduction Link

https://codepen.io/kdeds/pen/EaNjrbp?editors=0010

On the codepen, you can observe the bug easily with the axis label. The base config specifies an axis label for both axes. The responsive config overrides other options for the axes, but not the label. 
Below 600px, the X axis label remains visible. The Y axis label does not.
If you remove the responsive config entirely, or even just for the yaxis, the Y axis label is once again visible.
If you override the responsive yaxis label, the label changes properly below 600px.
If you change the chart plotOptions to horizontal: true (so that the x axis becomes the data axis), the xaxis config still merges properly.

EduardF1:
## Problem

When a responsive breakpoint is active, yaxis options defined in the base config are lost — only the keys explicitly set in 
esponsive[n].options.yaxis survive.

### Root cause

In Responsive.js, the active-breakpoint path is:

```js
newOptions = CoreUtils.extendArrayProps(config, res[i].options, w)
newOptions = Utils.extend(w.config, newOptions)
this.overrideResponsiveOptions(newOptions)
```

CoreUtils.extendArrayProps calls config.extendYAxis(options, w) which normalises the responsive yaxis snippet into a full array (applying all default values). Utils.extend then encounters two arrays for the yaxis key and — because arrays are not plain objects — simply replaces the entire array rather than deep-merging it. The base yaxis (which may contain user-set titles, min/max, tick counts, etc.) is completely discarded.

**Why does xaxis work fine?** xaxis is a plain object, so Utils.extend recurses into it and merges correctly. yaxis can be an array, which hits the non-merging path.

### Reproduction

See the codepen in #5211: base config sets axis labels on both axes; responsive config overrides other keys without touching the label. Below the breakpoint the **y-axis label disappears** (its base value was overwritten), but the x-axis label is preserved correctly.

## Fix

After the normal merge, re-merge each yaxis array entry individually using the **raw** user overrides (before default-expansion), so only the keys explicitly written in the responsive config override the base:

```js
newOptions = CoreUtils.extendArrayProps(config, res[i].options, w)
newOptions = Utils.extend(w.config, newOptions)
// Utils.extend does not deep-merge arrays, so a responsive yaxis array
// silently replaces the entire base yaxis.  Re-merge each entry with
// its base counterpart so only explicitly overridden keys change.
if (Array.isArray(w.config.yaxis) && res[i].options?.yaxis) {
  const responsiveYaxis = Array.isArray(res[i].options.yaxis)
    ? res[i].options.yaxis
    : [res[i].options.yaxis]
  newOptions = {
    ...newOptions,
    yaxis: w.config.yaxis.map((baseAxis, idx) =>
      Utils.extend(baseAxis, responsiveYaxis[idx] || {}),
    ),
  }
}
this.overrideResponsiveOptions(newOptions)
```

Fixes #5211


EduardF1:
Pushed [0ba210b](https://github.com/EduardF1/apexcharts.js/commit/0ba210b) which fixes a subtle bug in the first revision and adds unit coverage.

The first cut read the responsive yaxis from `res[i].options.yaxis` *after* `CoreUtils.extendArrayProps` had populated it with the full default yAxis schema. That made default-`undefined` keys (`min`, `max`, `title.text`, ...) and default-`false`/`0`/`null` keys (`opposite`, ...) look like explicit overrides and clobber the base axis values they were supposed to leave alone. So in the trivial case `responsive: [{ breakpoint: 1024, options: { yaxis: [{ labels: { show: false } }] } }]` the breakpoint also wiped `min`/`max`/`title.text`/...

This revision:

- Clones the user's responsive `yaxis` BEFORE `extendArrayProps` runs and uses that slim object as the source for the per-index merge.
- Adds a small `mergeYaxisOverride` helper that recursively skips `undefined` source values for safety.
- Adds two `Responsive` unit tests in `tests/unit/responsive.spec.js`:
  - **single-axis**: keys not re-declared in the responsive override (`min`, `max`, `tickAmount`, `title.text`) survive the breakpoint switch.
  - **multi-axis**: an empty `{}` slot leaves the corresponding base axis untouched, while a sibling slot picks up its override on top of its own base settings.

Full `npx vitest run tests/unit/` is green locally (50 files, 1418 passed).

EduardF1:
Just rebased on `main` to keep the branch fresh; revision 0ba210b is unchanged from the May 5 push (responsive yaxis deep-merge fix plus the two new `Responsive` unit tests). When a maintainer has a moment, this should be a focused review since the diff is contained to `Responsive.js` and the unit-test file. cc @junedchhipa

junedchhipa (approved):

PR #5247, opened 2026-08-02
Add cspNonce export option to fix CSP violation during export (#5146)

### LINKED ISSUE #5146: CSP Violation during Export (Strict Policy Incompatibility)
### Description
Applications enforcing a Strict Content Security Policy (CSP) (without 'unsafe-inline') are experiencing console errors and broken exports (SVG/PNG). This is caused by the dynamic injection of a <style> block inside a foreignObject during the export process, this behavior was recently updated in commit 9ce4d0c5f766de65f827e45f888b416e90ea44a9.

## Code
src/modules/Exports.js Lines: [78-80](https://github.com/apexcharts/apexcharts.js/blob/main/src/modules/Exports.js)
```js
<style type="text/css">
  ${exportStyles}
</style>
``` 

## Versions
- ApexCharts: ^5.3.6
- Vue3-ApexCharts: ^1.10.0
- Environment: Vue 3 (Composition API) with strict Content Security Policy.

### Steps to Reproduce

1. Use a Vue 3 application with vue3-apexcharts (or Vanilla ApexCharts).
2. Set a strict CSP via Meta tag: <meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self';">
3. Render a chart with a legend enabled.
4. Trigger the "Export to SVG" or "Export to PNG" via the toolbar or direct method call.
5. Check the Browser Console.

### Expected Behavior
The chart should export successfully without violating security policies. Styles should be applied via SVG attributes (e.g., fill, stroke, display) instead of an injected <style> string, or the library should provide a way to pass a nonce to the export styles.

### Actual Behavior
The browser blocks the style injection, throwing a CSP violation error, often resulting in unstyled exports or failed downloads.

### Screenshots

Image

## Evidence of Awareness (Commit 9ce4d0c)
The internal test file browser-sync-config.js confirms the developers were aware of this restriction, as they manually added a nonce to the test environment:

```js
const TEST_NONCE = '47ebaa88ef82ffb86e4ccb0...'
const cspPolicies = [`style-src 'self' 'nonce-${TEST_NONCE}'`]
```
However, the exported SVG and PNG does not inherit any application-level nonce, making this fix unavailable for end-users.

## Suggested Fixes
1. (Preferred): Move essential export styles from a <style> block to inline SVG attributes (e.g., fill="...", stroke="...", display="none") on the elements themselves. SVG attributes are not restricted by style-src 'unsafe-inline'.

2. (Opt-in): Provide a configuration flag (e.g., chart.export.ignoreStyles) to allow a "CSS-free" export for strict CSP environments, with default width and height etc...

3. (Nonce Support): Allow passing a cspNonce through the chart configuration to be injected into the generated <style> tag.

### Reproduction Link
https://codesandbox.io/p/sandbox/vue-basic-example-forked-3cw8zf?file=%2Findex.html%3A18%2C8

hello-himanshu-grover:
## What
Adds a `chart.toolbar.export.cspNonce` config option so a caller can
supply a CSP nonce that gets attached to the `<style>` tag injected
into SVG/PNG exports.

## Why
Closes #5146. Under a strict Content Security Policy (`style-src`
without `'unsafe-inline'`), that injected `<style>` tag is blocked by
the browser, breaking exports and throwing CSP violation errors in the
console — as reported with a reproduction and screenshots in the issue.

## Changes
- `src/modules/settings/Options.js`: new `export.cspNonce` default
  option (undefined by default, fully backward-compatible)
- `src/modules/Exports.js`: applies `nonce="..."` to the injected
  `<style>` tag in `getSvgString()` when `cspNonce` is set
- `types/apexcharts.d.ts`: matching `cspNonce?: string` type addition

## Usage
```js
new ApexCharts(el, {
  chart: { toolbar: { export: { cspNonce: 'your-per-request-nonce' } } }
})
```

## Testing
Syntax-checked both modified files locally. Wasn't able to run the
full suite in my environment due to a dependency (`@blazediff/core`)
resolving through a mirror my network didn't allow — happy to add
tests or address CI feedback if maintainers want a specific test shape.

### DIFF
--- a/src/modules/Exports.js
+++ b/src/modules/Exports.js
@@ -123,6 +123,12 @@ class Exports {
         exportStyles += apexchartsLegendCSS
       }
 
+      // A strict CSP (`style-src` without 'unsafe-inline') blocks this
+      // injected <style> tag unless it carries a nonce the page allow-lists.
+      // Let consumers pass that nonce through chart.toolbar.export.cspNonce.
+      const cspNonce = w.config.chart.toolbar.export.cspNonce
+      const nonceAttr = cspNonce ? ` nonce="${cspNonce}"` : ''
+
       let svgString = `
         <svg xmlns="http://www.w3.org/2000/svg"
           version="1.1"
@@ -133,7 +139,7 @@ class Exports {
           width="${w.globals.svgWidth}px" height="${w.globals.svgHeight}px">
           <foreignObject width="100%" height="100%">
             <div xmlns="http://www.w3.org/1999/xhtml" style="width:${width}px; height:${height}px;">
-            <style type="text/css">
+            <style type="text/css"${nonceAttr}>
               ${exportStyles}
             </style>
               ${serializedNode}
--- a/src/modules/settings/Options.js
+++ b/src/modules/settings/Options.js
@@ -627,6 +627,10 @@ export default class Options {
             },
             scale: undefined,
             width: undefined,
+            // A CSP `style-src` nonce to apply to the `<style>` tag injected
+            // into SVG/PNG exports, so exports work under a strict Content
+            // Security Policy that disallows 'unsafe-inline' styles.
+            cspNonce: undefined,
           },
           autoSelected: 'zoom', // accepts -> zoom, pan, selection, measure
         },
--- a/types/apexcharts.d.ts
+++ b/types/apexcharts.d.ts
@@ -1470,6 +1470,12 @@ type ApexChart = {
       }
       width?: number
       scale?: number
+      /**
+       * A CSP `style-src` nonce to apply to the `<style>` tag injected into
+       * SVG/PNG exports, so exports aren't blocked under a strict Content
+       * Security Policy that disallows 'unsafe-inline' styles.
+       */
+      cspNonce?: string
     }
     autoSelected?: 'zoom' | 'selection' | 'pan' | 'measure'
   }


junedchhipa:
Thanks for tackling this, and for the clear writeup! The direction is reasonable and nicely backward-compatible. A few things before we merge:

### 1. Please reuse the existing `chart.nonce` instead of adding a second option

We already ship a `chart.nonce` option (`Options.js`, typed in `apexcharts.d.ts`, tested in `tests/unit/nonce.spec.js`) that exists for exactly this CSP reason. It's applied to the injected stylesheets in `apexcharts.js` and `legend/Helpers.js`:

```js
const nonce = this.opts.chart?.nonce || this.w.config.chart.nonce
if (nonce) css.setAttribute('nonce', nonce)
```

The export `<style>` is just another injection point this nonce never happened to cover. Adding a separate `toolbar.export.cspNonce` means a user who already set `chart.nonce` will still get blocked exports until they discover and duplicate the value. Could we read `chart.nonce` here instead?

```js
const cspNonce = w.config.chart.toolbar.export.cspNonce || w.config.chart.nonce
```

Keeping `cspNonce` as an optional override is fine, but it should fall back to `chart.nonce`.

### 2. Can you verify it actually works in a strict-CSP browser?

Totally understand the local env issue. Two things make me want a real-browser confirmation before merging, because they may defeat the nonce:

- **Serialization round-trip.** `getSvgString()` builds the string, parses it via `DOMParser` (`svgStringToNode`), then re-serializes with `XMLSerializer`. Browser "nonce hiding" empties the `nonce` content attribute on nonceable elements (SVG `<style>` included), so the serialized output can end up as `nonce=""`. The existing `chart.nonce` path sets the attribute on a live element and never re-serializes, so it doesn't prove this path survives.
- **Where the violation fires.** The PNG path loads the SVG via `new Image()` + a `data:` URI, and the SVG path downloads a blob. Image-mode SVG and downloaded files generally aren't subject to the embedder's CSP. Could you confirm which export path throws in the repro, and that the nonce is present in the final serialized output on that path?

### 3. Minor: escape the value

`nonce="${cspNonce}"` is interpolated unescaped into the markup. Low risk since it's developer-supplied, but a value containing `"` would break the attribute. A small escape/validation would be good defensive hygiene.

### 4. Tests + docs

Would be great to add a test (`tests/unit/nonce.spec.js` is a good pattern to mirror) and a docs note. The option and type additions themselves look correct.

---

One bigger-picture note: the preferred fix in the issue (option #1) was to drop the injected `<style>` entirely by removing the hidden chrome nodes from the clone before serialization, since `exportStyles` is mostly just `display:none` on toolbar/tooltip elements plus the legend CSS. That would fix it for everyone with no config. The nonce is a fine incremental opt-in on top of that, but worth considering.

PR #5248, opened 2026-08-03
fix: respect series.hidden flag on initial render by emptying collapsed series data

### LINKED ISSUE #5213: "hidden" flag on series does not hide the series on render
### Description
Series that have the "hidden" flag set to true are still rendered in a line chart.
This bug appeared in version 5.10.4 (it works flawlessly in 5.10.3).

### Steps to Reproduce

1. Create a line chart with multiple series
2. Set the "hidden" flag on some series to false
3. Render the chart

### Expected Behavior
Only render the series that are not "hidden"

### Actual Behavior
You will see all series rendered in the chart. You will also see the hidden series' labels dimmed, which confirms that the "hidden" flag is present and that it does something.

### Screenshots

Image

### Reproduction Link
This shows multiple series. Half of them should be hidden, but all of them are rendered in the chart.
[https://codesandbox.io/p/devbox/react-basic-example-forked-czgd8h?workspaceId=ws_WJLdvWLykSPYYPCutRtykc](https://codesandbox.io/p/devbox/react-basic-example-forked-czgd8h?workspaceId=ws_WJLdvWLykSPYYPCutRtykc)

waterWang:
## Description

The `series.hidden` config option was being tracked in `collapsedSeriesIndices` global state via `getSeriesAfterCollapsing()`, but the actual series data was not being emptied before rendering. This caused hidden series to still be drawn (visible) on the chart, even though the legend labels correctly showed them as dimmed.

## Root Cause

When a user sets `hidden: true` on a series config, the `create()` method in `apexcharts.js` iterates through the series and calls `getSeriesAfterCollapsing()` for each hidden series. This correctly adds the index to `collapsedSeriesIndices`, but the cloned series data returned by the function is still passed to `parseData()` with all values intact. The rendering pipeline then draws the SVG paths for those series, and while the CSS class `apexcharts-series-collapsed` (opacity: 0) is added, the paths are still present in the DOM.

## Fix

Empty the data array for all indices tracked in `collapsedSeriesIndices` and `ancillaryCollapsedSeriesIndices` after the initial `forEach` loop. This follows the same pattern already used in `LegendHelpers._getSeriesBasedOnCollapsedState()`, which is called when a series is hidden/risen via legend interaction.

## Testing

1. Create a line chart with multiple series
2. Set `hidden: true` on one or more series
3. The hidden series should not be rendered on the chart

Fixes #5213

### DIFF
--- a/src/apexcharts.js
+++ b/src/apexcharts.js
@@ -326,6 +326,21 @@ export default class ApexCharts {
       }
     })
 
+    // Empty data for hidden/collapsed series so they are not rendered
+    if (
+      gl.collapsedSeriesIndices.length > 0 ||
+      gl.ancillaryCollapsedSeriesIndices.length > 0
+    ) {
+      series.forEach((s, si) => {
+        if (
+          gl.collapsedSeriesIndices.indexOf(si) > -1 ||
+          gl.ancillaryCollapsedSeriesIndices.indexOf(si) > -1
+        ) {
+          s.data = []
+        }
+      })
+    }
+
     const combo = CoreUtils.checkComboSeries(series, w.config.chart.type)
     gl.comboCharts = combo.comboCharts
     gl.comboBarCount = combo.comboBarCount


junedchhipa:
Thanks for the PR! 

I tried to reproduce this on current `main` (6.7.0) and I can't get the bug to trigger. With `hidden: true` set on one or more series at init, the hidden series already get the `apexcharts-series-collapsed` class, their `data` is already emptied, and no line/area/bar path is drawn.

The reason is that `create()`'s existing `hidden`-flag loop calls `getSeriesAfterCollapsing()`, which ends in `_getSeriesBasedOnCollapsedState()` and empties `data` for exactly the same collapsed/ancillary indices this PR targets. I applied your patch and the render output is byte-identical to `main` across every case I tried (single hidden, half-hidden, bar, `showAlways` yaxis, and legend-collapse followed by `updateOptions`). So the added block re-empties already-empty arrays.

I think #5213 (reported on 5.10.4) was fixed by other work since then.

1. Could you share a reproduction pinned to the **latest** version? The linked sandbox looks like it predates the fix.
2. One concern with the current approach: `create()` runs as `create(this.w.config.series, …)`, so when no series has a `hidden` flag, `series === w.config.series` and `s.data = []` mutates the user's config in place (the existing path deliberately works on a `Utils.clone`).

Issue #4216, opened 2024-02-07
Can't modify stroke width of multi-series sparkline chart


lmansur:
### Description

Passing a two-element array to `stroke.width` while using a two-series sparkline chart breaks the rendering of the chart. It seems related to this commit, specifically the part where math is being performed on `stroke.width`, even though it might be an array: https://github.com/apexcharts/apexcharts.js/commit/2ac1adba98c30b6c9cff3ff4811cc96d6f2cfcac

### Steps to Reproduce

1. Create a two-series sparkline chart
2. Add two values for `stroke.width`, like `[2, 3]`

### Expected Behavior

Render each series with the given stroke width

### Actual Behavior

The chart can't be rendered

### Screenshots

When not specifying the stroke width:
![image](https://github.com/apexcharts/apexcharts.js/assets/5934604/c5fa9fc0-e6b0-4586-b867-6e204f612c20)

When specifying a two-element array stroke width:
![image](https://github.com/apexcharts/apexcharts.js/assets/5934604/fff1ce01-4948-44ec-a158-bffd9cb973f2)

### Reproduction Link

https://codepen.io/lmansur/pen/mdojgWm?editors=1111

PR #5254, opened 2026-08-08
fix: remove extra bottom margin on sparkline charts (Closes #5137)

### LINKED ISSUE #5137: Extra bottom margin when using sparkline
### Description

When a chart is built with sparkline option, an extra margin (about 2px) is visible at the bottom.

### Steps to Reproduce

Just create a sparkline chart and put it inside a div with a border

### Expected Behavior

Remove extra margin

### Actual Behavior

See screenshot below

### Screenshots

Image

### Reproduction Link

https://codepen.io/dragonofmercy/pen/raeqYNm

waterWang:
Fixes #5137

## Summary
Removes the unused ~2px bottom margin on sparkline charts by:
1. Dropping the sparkline-specific grid padding that was set to `maxStrokeWidth / 2` — sparklines are meant to be tight, edge-to-edge, so automatic stroke-width padding is unnecessary
2. Changing the non-axis chart `offY` from 1 to 0 for sparklines (only affects pie/radial sparklines)

## Testing
- All 2131 unit tests pass
- Build succeeds (UMD, ESM, CommonJS, SSR)
- Manual verification: sparkline chart now fills its container without extra bottom space

### DIFF
--- a/src/modules/Core.js
+++ b/src/modules/Core.js
@@ -810,7 +810,7 @@ export default class Core {
     const isPercentHeight = heightStr.includes('%')
 
     let legendHeight = 0
-    let offY = w.config.chart.sparkline.enabled ? 1 : 15
+    let offY = w.config.chart.sparkline.enabled ? 0 : 15
     offY += w.config.grid.padding.bottom
 
     if (
--- a/src/modules/dimensions/Dimensions.js
+++ b/src/modules/dimensions/Dimensions.js
@@ -54,10 +54,6 @@ export default class Dimensions {
     this.lgRect = this.dimHelpers.getLegendsRect()
     this.datalabelsCoords = { width: 0, height: 0 }
 
-    const maxStrokeWidth = Array.isArray(w.config.stroke.width)
-      ? Math.max(...w.config.stroke.width)
-      : w.config.stroke.width
-
     if (this.isSparkline) {
       if (w.config.markers.discrete.length > 0 || w.config.markers.size > 0) {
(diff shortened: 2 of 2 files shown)


junedchhipa:
Thanks for digging into this, and sorry for the slow turnaround.

I measured the patch against `main` and I don't think it addresses the gap in #5137. The `maxStrokeWidth / 2` grid padding it removes is the stroke half-width reservation (added for #4216, `558cb45a0`), and since it equals exactly half the stroke width, deleting it doesn't reclaim any space: the canvas height and the visible bottom of the ink are unchanged. What does change is that the outer half of the stroke now falls outside the root SVG and gets clipped, so any point at the y-axis min/max loses half its line width. Our own sparklines demo (`Apex.stroke.width = 3`) and heatmap sparklines are affected.

The `offY` change also has a larger effect than 1px: `gridPad` aliases `w.config.grid.padding`, so `Core.js` was reading the stroke padding back out of config. Pie/donut/radial sparklines on auto height lose `1 + strokeWidth/2`, and PNG export inherits the same crop.

The actual bottom gap looks like it comes from y-axis nice-scale headroom or the `svgHeight`/`minHeight` math around `Core.js:708-736`, which this PR doesn't touch. Closing this one since a correct fix wouldn't share code with it, but #5137 stays open and a follow-up in that area would be very welcome. Anything that lands there should come with a test asserting `translateY >= strokeWidth / 2` - the current suite can't catch this class of regression.

PR #5255, opened 2026-08-08
fix(tooltip): clamp treemap tooltip within grid bounds on mobile

### LINKED ISSUE #5121: Tooltip exceeding screen limits in mobile screen
### Description

When using Treemap charts, the tooltips can exceed the screen limits. The tooltip is positioned based on the click on the rectangle, and depending where the click is, the tooltip will be displayed with some part out of the screen.

This is even observable in the examples in the official docs.

### Steps to Reproduce

1. Access the reproduction link
2. Set the browser responsive mode with minimum mobile screen size (320px)
3. Click the "Ahmedabad" rectangle in the right side of it, or the "Mumbai" rectangle, if using the official docs

### Expected Behavior

The tooltips should be limited to the screen surroundings.

### Actual Behavior

The tooltips are exceeding the screen limit on mobile sizes.

### Screenshots

Image

### Reproduction Link

[Reproduction link](https://codepen.io/gabsatoempiricus/pen/ZYQxXLp)
[Backup ](https://apexcharts.com/javascript-chart-demos/treemap-charts/basic/)

waterWang:
## Description

Fixes #5121

Treemap tooltips were positioned beside the hovered cell without any grid-bound clamping. On small viewports (e.g. 320px mobile), the tooltip overflowed the screen edge when the cell was near the right or bottom of the chart.

## Changes

- **`src/modules/tooltip/Intersect.js`** — Added grid-bound clamping to `handleHeatTreeTooltip`'s legacy treemap path, mirroring the existing bar-tooltip behavior. Clamps `x` to `[0, gridWidth - ttW]` and `y` to `[0, gridHeight - ttH]` (elWrap-local coords).
- **`tests/unit/tooltip-positioning.spec.js`** — Added 4 tests covering right-edge, left-edge, top-edge, and bottom-edge clamping.

## Testing

- All 2141 unit tests pass (including 4 new ones).
- Manual verification: with `gridWidth=320, gridHeight=400`, a cell at `cx=280` previously produced `x=380` (overflowing by 60px); now clamped to `x=200` (right edge of grid).

### DIFF
--- a/src/modules/tooltip/Intersect.js
+++ b/src/modules/tooltip/Intersect.js
@@ -194,6 +194,15 @@ class Intersect {
         (y > w.layout.gridHeight / 2 ? ttCtx.tooltipRect.ttHeight : 0)
     }
 
+    // Clamp tooltip within grid bounds so it doesn't overflow the
+    // screen on small viewports (fixes treemap tooltip on mobile).
+    const ttW = ttCtx.tooltipRect.ttWidth || 0
+    const ttH = ttCtx.tooltipRect.ttHeight || 0
+    if (x < 0) x = 0
+    if (x + ttW > w.layout.gridWidth) x = w.layout.gridWidth - ttW
+    if (y < 0) y = 0
+    if (y + ttH > w.layout.gridHeight) y = w.layout.gridHeight - ttH
+
     return {
       x,
       y,
--- a/tests/unit/tooltip-positioning.spec.js
+++ b/tests/unit/tooltip-positioning.spec.js
@@ -9,6 +9,7 @@
  */
 import { describe, it, expect, vi } from 'vitest'
 import TooltipPosition from '../../src/modules/tooltip/Position.js'
+import Intersect from '../../src/modules/tooltip/Intersect.js'
 import { renderMarkerSVG } from '../../src/modules/tooltip/Marker.js'
 import { createChartWithOptions } from './utils/utils.js'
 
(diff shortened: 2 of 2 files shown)


junedchhipa (changes requested):
Thanks for this, and for the clear before/after numbers in the description.

The overflow is real and the patch does close #5121. I measured it in Chromium at a 320px viewport
with a treemap using longer category labels: on `main` the tooltip escapes the chart box by 89px to
the right, 64px to the left and 26px below; with this patch every edge lands inside the box. Three
things to sort out before it can land.

### 1. One of the four new tests fails, on this branch

`clamps y to 0 when tooltip would overflow the top edge in followCursor mode` fails with
`expected 10 to be +0`. I ran the spec in a worktree at `79beaeb4c` to rule out a rebase artifact,
and it fails there too, so the "all 2141 unit tests pass (including 4 new ones)" claim does not
hold as authored (`Intersect.js` has not changed on `main` since your base commit).

The cause is in the test's arithmetic, not the fix. The `- ttH` term in the followCursor branch is
conditional:

```js
y = (w.interact.clientY ?? 0) - seriesBound.top -
    (y > w.layout.gridHeight / 2 ? ttCtx.tooltipRect.ttHeight : 0)
```

With `cy = 100` and `height = 30` the cell-derived `y` lands in the top half, so nothing is
subtracted, `y` is `10`, and no clamp is needed. The test never exercises a top-edge clamp. The
other three do guard the fix, for what it is worth: all four fail when I revert the source hunk.

### 2. The clamp compares an elWrap coordinate against a grid-local length

`x` and `y` here are elWrap-local: they are written straight to `tooltipEl.style.left/top`, and the
followCursor branch builds them as `clientX - elWrap.left`. But the bounds are `gridWidth` and
`gridHeight`, which are grid *lengths*. The grid's right edge in elWrap coords is
`translateX + gridWidth`. Treemap gets away with it because `translateX` is 0 there, which is why
the reported case looks clean, but any chart whose grid is inset by axis labels is over-clamped by
exactly `translateX`.

Measured on a 600px-wide heatmap with `followCursor: true` (which takes this same legacy path):
`translateX = 124`, `gridWidth = 456`, chart right edge at 600.

| cursor x | tooltip right edge, `main` | tooltip right edge, this PR |
|---------:|---------------------------:|----------------------------:|
| 454 | 454 | 454 |
| 494 | 494 | 456 |
| 534 | 534 | 456 |
| 574 | 534 | 456 |

From `cursorX ≈ 494` the tooltip stops tracking the pointer and ends up as much as 274px to its
left, with 144px of chart still unused to the right. The arrow-mode heatmap branch a few lines
above already handles this properly by converting to elWrap coords first
(`gridOffsetXInElWrap`, `w.layout.translateY`); reusing those two offsets for the bounds here would
keep both paths consistent.

### 3. Clamp order lets the tooltip escape to the left

The left clamp runs before the right clamp, so when the tooltip is wider than the grid the second
assignment wins and `x` ends up negative:

```js
if (x < 0) x = 0
if (x + ttW > w.layout.gridWidth) x = w.layout.gridWidth - ttW   // can go negative
```

On a 200px-wide treemap with long labels (`ttW = 244`, `gridWidth = 200`) this puts
`style.left: -43.8px`, i.e. 44px of tooltip left of the chart, where `main` had no left overflow at
all. A single clamped expression avoids the ordering question entirely:

```js
x = Math.max(gridLeft, Math.min(x, gridRight - ttW))
y = Math.max(gridTop, Math.min(y, gridBottom - ttH))
```

### Smaller notes

- The branch you are editing is shared with heatmap (`arrow: false` or `followCursor: true`), not
  treemap only. Clamping those too is a fine outcome, but it is worth saying so in the description,
  and it is what makes (2) a user-visible regression rather than a theoretical one.
- The new tests hand-build `w` and `ttCtx` mocks with `translateX: 0`, which is precisely why they
  cannot see (2). `tests/interaction/specs/` runs real Chromium and already has a heatmap tooltip
  spec (`heatmap-tooltip-arrow.spec.js`) to model a treemap one on; a spec that asserts the tooltip
  rect stays inside the chart rect at a 320px viewport would be a much stronger guard than the
  mocks, and would survive future refactors of the placement math.

Happy to take another look once the bounds are in elWrap space and the failing test is sorted.

junedchhipa:
Thank you for this, and I am sorry it sat for a month. You reported and fixed this before anyone else, and it is not good that a later PR for the same issue went in first. That is on me, not on you.

#5290 landed for a specific reason worth knowing: the `x` here ends up as the tooltip element's `left`, and that element is parented to the chart wrapper rather than the grid, so its origin is the wrapper edge and not the plot edge. Clamping to `0..gridWidth` is the right idea against the wrong origin, and it leaves the box short of the plot by the grid offset at both ends. The merged version measures the grid's offset inside the wrapper and clamps against that.

#5121 is closed, so I am closing this too, with the credit for finding it yours. Your other open PRs are the next thing I am going through, and I will not let them go stale the way this one did.

PR #5257, opened 2026-08-08
fix: make SVG export CSP-safe by inlining styles instead of injecting <style> tag

### LINKED ISSUE #5146: CSP Violation during Export (Strict Policy Incompatibility)

waterWang:
## Description

Applications enforcing a strict Content Security Policy (CSP) without `'unsafe-inline'` experience console errors and broken SVG/PNG exports because the export process injects a `<style type="text/css">` block inside a `<foreignObject>` in the generated SVG string.

This PR replaces the injected `<style>` block with direct DOM manipulation on the cloned node before serialization:

1. **Transient UI overlays** (`.apexcharts-tooltip`, `.apexcharts-toolbar`, `.apexcharts-xaxistooltip`, `.apexcharts-yaxistooltip`, `.apexcharts-xcrosshairs`, `.apexcharts-ycrosshairs`, `.apexcharts-zoom-rect`, `.apexcharts-selection-rect`) are hidden via `element.style.display = 'none'` on the cloned node.

2. **Legend layout styles** are applied as inline `style` attributes on the legend elements (`.apexcharts-legend`, `.apexcharts-legend-series`, `.apexcharts-legend-text`, `.apexcharts-legend-marker`, `.apexcharts-inactive-legend`, `.apexcharts-hidden-zero-series`, `.apexcharts-hidden-null-series`).

3. The `<style>` block and the `apexchartsLegendCSS` import are removed since they are no longer needed.

## Benefits

- ✅ No CSP console errors during export
- ✅ SVG/PNG export works under strict CSP policies
- ✅ All 2137 existing tests pass (94 test files)
- ✅ No behavioral change for applications without CSP
- ✅ Smaller exported SVG (no inline `<style>` block)

## Closes

Fixes #5146


junedchhipa:
Thanks for this: the underlying problem is real and worth fixing. Serializing a `<style>` block into the exported SVG breaks export for anyone on a strict CSP, and nobody had picked it up.

Reviewing it, though, the rewrite traded the stylesheet for a lossy subset of it, so I found a few things that needed sorting before it could land. Recording them here so the reasoning is public:

**The CSP goal wasn't actually met in the default config.** `legend/Helpers.appendToForeignObject()` injects the legend stylesheet into `elLegendForeign`, which is a descendant of `elWrap`. Since the export clones `elWrap`, that `<style>` was cloned and serialized regardless. With `chart.injectStyleSheet` at its default, the exported SVG still contained an inline `<style>` and still tripped a strict CSP. The clone needs the `<style>` stripped.

**`querySelector` hid only the first match per selector, where the CSS hid all of them.** A chart can hold more than one of those overlays: one `.apexcharts-yaxistooltip` per y-axis (2 for a 2-axis chart), and point annotations add a second `.apexcharts-tooltip`. Their `opacity: 0` came from the page stylesheet, which the export no longer carries, so the leftovers rendered visibly in the export.

**Hardcoding `font-size: 14px` overrode the configured size.** `Legend.js` sets `legend.fontSize` inline (default 12px), and the legend box is measured at that size. With `legend.fontSize: '20px'` the export rendered the text at 14px inside a box sized for 20px. Same class of problem hit the heatmap gradient legend: it deliberately sets `display: block; overflow: visible; padding: 0` on the legend wrap, and the unconditional wrap styling put back the clipping and scrollers that override exists to prevent.

The root cause of those last two is that inline styles set by a module have to keep winning over what the stylesheet used to provide. The fix applies each property only when the element has no inline value for it, which mirrors the cascade the stylesheet sat in.

**The inlined subset was missing layout rules**, notably `flex-wrap: wrap`, `flex-direction: column` for side and grouped-horizontal legends, the alignment rules, `.apexcharts-legend-group { display: flex }`, and marker `position: relative`. Also `.apexcharts-flip-x/-y`, which rounded stacked bars rely on. With `injectStyleSheet: false` (exactly what a strict-CSP app sets) there was no stylesheet left to fall back on, so a left/right legend exported as one horizontal row and bottom legends stopped wrapping: a regression against the old export.

Smaller: the `legend.show && children.length > 0` guard was dropped, and the rewrite added 17 `TS2339` errors to this `@ts-check` file (`.style` does not exist on `Element`).

I've pushed the fixes to this branch rather than sending it back to you, since most of them turn on internals that aren't obvious from outside the codebase (which foreignObject the stylesheet lands in, why the gradient legend overrides those exact properties, what the full legend ruleset covers). Your commit stays in the history. I also added `tests/unit/export-styles.spec.js` to pin the behaviour, since this is the kind of fidelity that regresses silently: I verified each of the three main issues above turns that suite red when reintroduced.

One incidental finding, not yours and not fixed here: `getSvgString()` returns a `<parsererror>` under jsdom, because jsdom's XMLSerializer emits `xmlns` on the `<svg>` root in addition to the one svg.js already set, and the re-parse rejects the duplicate attribute. It reproduces on `main` and doesn't affect real browsers, but it does mean unit tests can't assert on that return value. Worth a separate look.

Merging now. Thanks again for finding and chasing this one.

junedchhipa (approved):
Approach is right and the underlying CSP problem is real. Fidelity gaps addressed in f38ae33e, with tests/unit/export-styles.spec.js pinning the behaviour. Green against the fixed main.

PR #5258, opened 2026-08-08
fix: prevent zoom selection dragging from triggering chart click event

### LINKED ISSUE #5060: Selection zoom rectangle mouse up triggers click event
### Description

With xy zoom enabled and by dragging the selection zoom rectangle the click event is invoked on mouse up.

### Expected Behavior

Doing a zoom by dragging the zoom rectangle should not trigger the click event.

### Actual Behavior

Doing a zoom by dragging the zoom rectangle triggers the click event

### Reproduction Link

waterWang:
## Description

When a user performs a zoom selection by dragging a rectangle on the chart (with `chart.zoom.enabled: true` and `chart.zoom.type: 'xy'`), the `mouseup` event at the end of the drag was incorrectly firing the chart's `click` event handler (`chart.events.click` and the `click` event).

## Root Cause

The `Events.js` `setupEventHandlers` method listens for `mouseup` on the chart element and fires the click event callback. At the same time, `ZoomPanSelection.js` `svgMouseEvents` also handles `mouseup` for zoom/pan selection. Since the SVG element (where zoomPanSelection listens) is a child of the chart element, `mouseup` bubbles up — but there was no flag to distinguish a drag from a simple click.

## Fix

Two small changes:

1. **ZoomPanSelection.js** (`handleMouseUp`): Set `w.interact.wasDragged = this.dragged` before resetting the drag state, so the Events handler can check whether the mouseup concluded a drag. Also reset `wasDragged` on the next `mousedown` to prevent stale flags from a `mouseleave` abort.

2. **Events.js** (`setupEventHandlers`): Skip the click event callback when `w.interact.wasDragged` is true, then reset it.

## Testing

All 2137 tests pass across 94 test files.

Closes #5060

### DIFF
--- a/src/modules/Events.js
+++ b/src/modules/Events.js
@@ -136,6 +136,11 @@ export default class Events {
             (e.type === 'mouseup' && e.which === 1) ||
             e.type === 'touchend'
           ) {
+            // Skip click event if the mouseup was part of a drag (zoom/pan selection)
+            if (w.interact.wasDragged) {
+              w.interact.wasDragged = false
+              return
+            }
             if (typeof w.config.chart.events.click === 'function') {
               w.config.chart.events.click(e, me, opts)
             }
--- a/src/modules/ZoomPanSelection.js
+++ b/src/modules/ZoomPanSelection.js
@@ -252,6 +252,7 @@ export default class ZoomPanSelection extends Toolbar {
       this.startY = this.clientY - gridRectDim.top
 
       this.dragged = false
+      this.w.interact.wasDragged = false
       this.w.interact.mousedown = true
     }
 
@@ -320,6 +321,7 @@ export default class ZoomPanSelection extends Toolbar {
       this.hideSelectionRect(this.selectionRect)
     }
 
+    this.w.interact.wasDragged = this.dragged
     this.dragged = false
     this.w.interact.mousedown = false
   }


junedchhipa (changes requested):
Thanks for the write-up, the root-cause analysis is spot on: the mouseup does bubble from
`.apexcharts-svg` to `clickableArea`, and since `ZoomPanSelection` listens on the child it always
runs before `Events`, so the flag is set before it is read. I confirmed the fix works: on `main` a
150px drag-zoom fires `chart.events.click` once, and with this patch it fires zero times while
`zoomed` still fires. Unit suite stays green.

Two blockers and a test request.

### 1. Clicks that move a few pixels now fire nothing at all

`this.dragged` is set by *any* `mousemove` with the button held:

```js
if ((e.type === 'mousemove' && e.which === 1) || e.type === 'touchmove') {
  this.dragged = true
```

but the library's own definition of a drag, in `selectionDrawn`, has a threshold:

```js
if (me.dragged && (me.dragX > 10 || me.dragY > 10) && xLowestValue !== xHighestValue) {
```

So between 1px and 9px of movement the gesture is too small to zoom, and now also too "dragged" to
click. The user gets neither. Measured (mousedown, move N px, mouseup), clicks fired:

| gesture | `main` | this PR |
|---|---:|---:|
| click, no movement | 1 | 1 |
| click, 1px | 1 | 0 |
| click, 5px | 1 | 0 |
| click, 9px | 1 | 0 |
| drag-zoom, 150px | 1 (the bug) | 0 |

Touch is the case I would worry about most, since a real tap almost always jitters. A synthetic tap
with 3px of movement between `touchstart` and `touchend`: 1 click on `main`, 0 with this patch.

Reusing the existing threshold fixes it:

```js
this.w.interact.wasDragged =
  this.dragged && (this.dragX > 10 || this.dragY > 10)
```

`dragX`/`dragY` are already computed a few lines above in `handleMouseUp`. I built that variant and
re-ran the whole matrix: the drag-zoom click stays suppressed, the 1/5/9px clicks and the 3px tap
all fire again, and a 40px tap stays suppressed. That seems like the behaviour you want, and it has
the nice property that "what counts as a drag" is defined once.

### 2. This breaks `tsc`

`wasDragged` is not declared on `InteractState` (`src/types/internal.d.ts:66`), so the patch adds
four errors:

```
src/modules/Events.js(140,28): error TS2339: Property 'wasDragged' does not exist on type 'InteractState'.
src/modules/Events.js(141,26): error TS2339: Property 'wasDragged' does not exist on type 'InteractState'.
src/modules/ZoomPanSelection.js(255,23): error TS2339: Property 'wasDragged' does not exist on type 'InteractState'.
src/modules/ZoomPanSelection.js(324,21): error TS2339: Property 'wasDragged' does not exist on type 'InteractState'.
```

Add `wasDragged?: boolean` next to `shiftWasPressed?: boolean`, which lives there for the same
reason: state that has to survive the destroy-and-recreate that every `_updateOptions` performs.
(Heads up that `npm run typecheck` is currently red on `main` for unrelated reasons, so these
would not have stood out locally.)

### 3. Please add an interaction test

jsdom cannot express this, but `tests/interaction/specs/` runs real Chromium and
`zoom-pan-selection.spec.js` already has a `dragOnChart` helper. Two assertions would lock both
halves down: a drag-zoom fires `zoomed` and no `click`, and a small-movement click still fires
`click`.

### Checked, and not problems

I went looking for stale-flag paths and did not find any, so no need to worry about these:

- Plain hover movement does not set `dragged`: Chrome reports `which: 0` for a button-less
  `mousemove`, so `w.interact.wasDragged` stays `false` after hovering across the plot and leaving
  it.
- A drag aborted by leaving the SVG (so `mouseleave` runs `handleMouseUp` with no matching mouseup)
  does leave the flag set, but the very next `mouseup` reaching `clickableArea` consumes and resets
  it. I verified that legend clicks after an aborted drag fire normally, then keep firing.
- Element-level handlers are untouched: `dataPointSelection`, legend clicks and axis-label clicks
  are bound on the elements themselves, so only the public chart-level `click` is gated.

One behaviour note for the description: pan-mode drags are suppressed too. That looks intended, but
it is a change to `chart.events.click` semantics beyond the zoom case in #5060, so it is worth
stating explicitly.

PR #5260, opened 2026-08-08
fix: include brush autoScaleYaxis boundary points within one rendered pixel (#5251)

### LINKED ISSUE #5251: Brush selection may exclude a boundary point from autoScaleYaxis due to subpixel timestamp rounding
### Description

When a datetime chart is controlled by a brush chart, `autoScaleYaxis` may calculate a Y range that does not include the rightmost rendered point.

As a result, the line can leave the visible Y range or be clipped at the top of the chart.

The same X window scales correctly when it is reached by panning the main chart.

This appears to happen because brush selection coordinates go through a DOM pixel → timestamp conversion. Subpixel rounding can produce an `xaxis.max` value that is slightly smaller than the timestamp of the boundary data point.

The point is then excluded from the Y extrema calculation, although the line segment leading to that point is still rendered inside the plot area.

### Version

- ApexCharts: `6.7.0`
- vue3-apexcharts: `1.11.1`, although the issue appears to be in ApexCharts itself
- Chart types: datetime area chart controlled by a bar brush chart
- Browser: Chrome

### Configuration

Both relevant auto-scale options are enabled:

```js
// Main chart
chart: {
  type: "area",
  zoom: {
    enabled: true,
    type: "x",
    autoScaleYaxis: true,
  },
}

// Brush chart
chart: {
  type: "bar",
  brush: {
    enabled: true,
    target: "main-chart",
    autoScaleYaxis: true,
  },
}
```

### Minimal reproduction

```html
<div id="main"></div>
<div id="brush"></div>

<script src="https://cdn.jsdelivr.net/npm/apexcharts"></script>

<script>
  const DAY = 24 * 60 * 60 * 1000
  const start = Date.UTC(2026, 5, 15)

  const values = [
    483, 438, 438, 438, 438,
    390, 388, 437, 448, 448,
    450, 447, 540,
  ]

  const data = values.map((y, index) => [
    start + index * DAY,
    y,
  ])

  const initialMin = data[5][0]
  const initialMax = data[data.length - 1][0]

  const main = new ApexCharts(document.querySelector("#main"), {
    chart: {
      id: "main-chart",
      type: "area",
      height: 240,
      animations: {
        enabled: false,
      },
      zoom: {
        enabled: true,
        type: "x",
        autoScaleYaxis: true,
      },
      toolbar: {
        autoSelected: "pan",
      },
    },
    series: [{
      name: "Price",
      data,
    }],
    stroke: {
      curve: "smooth",
      width: 3,
    },
    markers: {
      size: 4,
    },
    xaxis: {
      type: "datetime",
      min: initialMin,
      max: initialMax,
    },
  })

  await main.render()

  const brush = new ApexCharts(document.querySelector("#brush"), {
    chart: {
      type: "bar",
      height: 100,
      animations: {
        enabled: false,
      },
      brush: {
        enabled: true,
        target: "main-chart",
        autoScaleYaxis: true,
      },
      selection: {
        enabled: true,
        xaxis: {
          min: initialMin,
          max: initialMax,
        },
      },
    },
    series: [{
      name: "Price",
      data,
    }],
    xaxis: {
      type: "datetime",
    },
    yaxis: {
      show: false,
    },
  })

  await brush.render()
</script>
```

### Steps to reproduce

1. Render the main chart and the brush chart.
2. Move or resize the brush selection.
3. Position its right edge at the last data point.
4. Observe the Y axis of the main chart.
5. Compare it with the result obtained by panning the main chart to the same X window.

The issue is easier to reproduce with fractional container widths, browser zoom other than 100%, or device pixel ratios that cause fractional SVG coordinates.

### Expected behavior

All points and line segments rendered inside the selected X window should participate in `autoScaleYaxis`.

The rightmost high-value point should increase the Y-axis maximum so that the line remains inside the plot.

Brush selection and main-chart pan should produce the same Y-axis range for the same effective X window.

### Actual behavior

After changing the X window through the brush:

- the rightmost point may be excluded from Y-axis extrema;
- the line leading to it is still visible;
- the line can leave the displayed Y range or be clipped at the top.

Panning the main chart recalculates the Y range correctly.

### Suspected cause

In `ZoomPanSelection`, the selection boundary is reconstructed from the selection rectangle using its DOM bounds:

```js
const selectionRect = this.selectionRect.node.getBoundingClientRect()
const relRight = this._screenXToPlotPx(selectionRect.right)
const maxX = AxisMapping.pxToDataX(w, relRight)
```

The pixel round-trip can return something similar to:

```js
selectionMax = lastPointTimestamp - smallRoundingError
```

Later, `Range.getMinYMaxY()` uses a strict comparison when selecting points for Y auto-scaling:

```js
while (
  lastXIndex > firstXIndex &&
  seriesX[lastXIndex] > config.xaxis.max
) {
  lastXIndex--
}
```

Therefore, even a very small timestamp difference excludes the last point.

Relevant source files:

- [`src/modules/ZoomPanSelection.js`](https://github.com/apexcharts/apexcharts.js/blob/main/src/modules/ZoomPanSelection.js)
- [`src/modules/Range.js`](https://github.com/apexcharts/apexcharts.js/blob/main/src/modules/Range.js)
- [`src/modules/AxisMapping.js`](https://github.com/apexcharts/apexcharts.js/blob/main/src/modules/AxisMapping.js)

### Suggested regression test

A regression test could simulate a brush boundary that differs from the last point by less than one rendered pixel in the X domain:

```js
it("includes a rendered brush boundary point in autoScaleYaxis", async () => {
  const lastX = Date.UTC(2026, 7, 3)
  const lastY = 540

  // Represents a DOM/SVG subpixel round-trip.
  const brushMax = lastX - subpixelTimestampError

  await applyBrushRange({
    min: lastX - 7 * DAY,
    max: brushMax,
  })

  const onePixelInDomain =
    (chart.w.globals.maxX - chart.w.globals.minX) /
    chart.w.layout.gridWidth

  expect(lastX - brushMax).toBeLessThan(onePixelInDomain)

  // The boundary segment/point is rendered, so it must fit the Y scale.
  expect(chart.w.globals.yAxisScale[0].niceMax)
    .toBeGreaterThanOrEqual(lastY)
})
```

An integration-level variation could:

1. Render a brush with a fractional CSS width.
2. Move its selection rectangle through the normal pointer interaction.
3. Read the resulting target X range.
4. Assert that any data point rendered within one pixel of the boundary is included in the calculated Y range.

### Possible fixes

Possible approaches could be:

1. Snap brush selection boundaries to an exact series X value when they are within a render-pixel tolerance.
2. Apply a pixel-domain tolerance when determining the first and last data indices for `autoScaleYaxis`.
3. Calculate Y extrema from the points/segments that are actually rendered after clipping.

The tolerance should probably be derived from the current X-domain-to-pixel ratio rather than using a fixed timestamp epsilon.

### Current workaround

Before calling `zoomX()`, we expand the brush range to the nearest non-null data points:

```js
function expandRangeToDataPoints(series, range) {
  let min = range.min
  let max = range.max
  let previousX
  let nextX

  for (const item of series) {
    for (const [x, value] of item.data) {
      if (value == null) continue

      if (x <= range.min && (previousX == null || x > previousX)) {
        previousX = x
      }

      if (x >= range.max && (nextX == null || x < nextX)) {
        nextX = x
      }
    }
  }

  return {
    min: previousX ?? min,
    max: nextX ?? max,
  }
}
```

This prevents the boundary point from being excluded, but it would be preferable for brush-controlled charts to handle this consistently inside ApexCharts.

Image

waterWang:
## Summary

Fixes #5251 — When a datetime chart is controlled by a brush chart with `autoScaleYaxis` enabled, the Y-axis range may exclude the rightmost (or leftmost) boundary data point.

### Problem

The brush selection boundary is reconstructed from the selection rectangle using its DOM pixel bounds (`selectionRect.node.getBoundingClientRect()` → `AxisMapping.pxToDataX`). The pixel→timestamp round-trip can produce a value slightly smaller than the exact data point timestamp (e.g. `lastPointTimestamp - 1ms`).

In `Range.getMinYMaxY()`, when selecting points for Y auto-scaling, the loop uses a **strict** comparison:

```js
while (
  lastXIndex > firstXIndex &&
  seriesX[lastXIndex] > config.xaxis.max   // <-- strict
) {
  lastXIndex--
}
```

So even a tiny sub-pixel rounding error causes the boundary point to be **excluded** from the Y extrema calculation, even though the line segment leading to it is still rendered inside the plot area. As a result the line can leave the visible Y range or be clipped.

Brush selection and main-chart panning produce different Y ranges for the same effective X window.

### Fix

Compute a one-rendered-pixel tolerance in the X domain and apply it to both boundary comparisons:

```js
const onePixelInDomain = (gl.maxX - gl.minX) / this.w.layout.gridWidth
// min side:  seriesX[firstXIndex] < cnf.xaxis.min - onePixelInDomain
// max side:  seriesX[lastXIndex]  > cnf.xaxis.max + onePixelInDomain
```

A data point whose timestamp differs from the brush boundary by less than one rendered pixel is now treated as being inside the window, matching what is actually rendered. Points far outside the range are still excluded (no false positives).

The tolerance is derived from the live X-domain-to-pixel ratio, as suggested in the issue, rather than a fixed timestamp epsilon.

### Verification

- New regression test: `brush-autoscaleyaxis-boundary.spec.js`
  - Includes a boundary point within one pixel of the brush max in `autoScaleYaxis`
  - Confirms a point far outside the range is still excluded (no false positive)
- Full unit suite: **1978 tests pass** (the 3 pre-existing `No such built-in module: node:` failures are unrelated to this change)

### DIFF
--- a/src/modules/Range.js
+++ b/src/modules/Range.js
@@ -102,11 +102,21 @@ class Range {
       let lastXIndex = series[i].length - 1
       if (autoScaleYaxis) {
         // Scale the Y axis to the min..max within the possibly zoomed X axis domain.
+        // Subpixel rounding in brush selection may produce a pixel→timestamp
+        // conversion that is slightly off exact data point timestamps, causing
+        // boundary points to be excluded from the Y range. (#5251)
+        let onePixelInDomain = 0
+        if (
+          gl.maxX !== gl.minX &&
+          this.w.layout.gridWidth > 0
+        ) {
+          onePixelInDomain = (gl.maxX - gl.minX) / this.w.layout.gridWidth
+        }
         if (cnf.xaxis.min) {
           for (
             ;
             firstXIndex < lastXIndex &&
-            this.w.seriesData.seriesX[i][firstXIndex] < cnf.xaxis.min;
+            this.w.seriesData.seriesX[i][firstXIndex] < cnf.xaxis.min - onePixelInDomain;
             firstXIndex++
           ) {
             // Intentionally empty - just incrementing firstXIndex
@@ -116,7 +126,7 @@ class Range {
           for (
             ;
             lastXIndex > firstXIndex &&
-            this.w.seriesData.seriesX[i][lastXIndex] > cnf.xaxis.max;
+            this.w.seriesData.seriesX[i][lastXIndex] > cnf.xaxis.max + onePixelInDomain;
             lastXIndex--
           ) {
             // Intentionally empty - just decrementing lastXIndex
--- a/tests/unit/brush-autoscaleyaxis-boundary.spec.js
+++ b/tests/unit/brush-autoscaleyaxis-boundary.spec.js
@@ -1,0 +1,136 @@
+import { createChartWithOptions } from './utils/utils.js'
+
+const DAY = 24 * 60 * 60 * 1000
+
+describe('Brush autoScaleYaxis boundary inclusion (#5251)', () => {
+  beforeEach(() => {
+    window.Apex = {}
+  })
+  afterEach(() => {
+    document.getElementsByTagName('html')[0].innerHTML = ''
+  })
+
+  it('includes a boundary point within one rendered pixel of the brush max', async () => {
+    const start = Date.UTC(2026, 5, 15)
+
+    // 12 data points — the last one has the highest Y value (540)
+    const data = [
+      [start + 0 * DAY, 483],
+      [start + 1 * DAY, 438],
+      [start + 2 * DAY, 438],
+      [start + 3 * DAY, 438],
+      [start + 4 * DAY, 438],
+      [start + 5 * DAY, 390],
+      [start + 6 * DAY, 388],
+      [start + 7 * DAY, 437],
+      [start + 8 * DAY, 448],
+      [start + 9 * DAY, 448],
+      [start + 10 * DAY, 450],
+      [start + 11 * DAY, 540],
+    ]
+
+    const lastX = data[data.length - 1][0]
+    const lastY = data[data.length - 1][1]
+
+    // The brush selection max is slightly before the last data point,
+    // simulating a subpixel rounding error in the DOM pixel→timestamp conversion.
+    const subpixelTimestampError = 1 // 1 ms before the last point
+    const brushMax = lastX - subpixelTimestampError
+
+    const chart = createChartWithOptions({
+      chart: {
+        id: 'main-chart',
+        type: 'area',
+        height: 240,
+        width: '800px',
+        zoom: {
+          enabled: true,
+          type: 'x',
+          autoScaleYaxis: true,
+        },
+        toolbar: {
+          autoSelected: 'pan',
+        },
+      },
+      series: [
+        {
+          name: 'Price',
+          data,
+        },
+      ],
+      stroke: {
+        curve: 'smooth',
+        width: 3,
+      },
+      markers: {
+        size: 4,
+      },
+      xaxis: {
+        type: 'datetime',
+        min: data[0][0],
+        max: lastX,
+      },
+    })
+
+    // Simulate a brush selection that ends slightly before the last data point
+    chart.zoomX(data[0][0], brushMax)
+
+    // The Y-axis max should include the last point (540) even though
+    // brushMax is slightly before lastX, because the difference is
+    // less than one rendered pixel in the X domain.
+    expect(chart.w.globals.yAxisScale[0].niceMax).toBeGreaterThanOrEqual(lastY)
+  })
+
+  it('excludes a point far outside the brush range (no false positive)', async () => {
+    const start = Date.UTC(2026, 5, 15)
+
+    const data = [
+      [start + 0 * DAY, 100],
+      [start + 1 * DAY, 200],
+      [start + 2 * DAY, 300],
+      [start + 3 * DAY, 400],
+      [start + 4 * DAY, 500],
+    ]
+
+    const chart = createChartWithOptions({
+      chart: {
+        id: 'main-chart',
+        type: 'area',
+        height: 240,
+        zoom: {
+          enabled: true,
+          type: 'x',
+          autoScaleYaxis: true,
+        },
+        toolbar: {
+          autoSelected: 'pan',
+        },
+      },
+      series: [
+        {
+          name: 'Price',
+          data,
+        },
+      ],
+      stroke: {
+        curve: 'smooth',
+        width: 3,
+      },
+      markers: {
+        size: 4,
+      },
+      xaxis: {
+        type: 'datetime',
+        min: data[0][0],
+        max: data[data.length - 1][0],
+      },
+    })
+
+    // Zoom to first 2 points only — the last point (500) is far outside
+    chart.zoomX(data[0][0], data[1][0])
+
+    // The Y-axis max should be based on the max of the first 2 points (200),
+    // not the last point (500) which is far outside the range.
+    expect(chart.w.globals.yAxisScale[0].niceMax).toBeLessThan(500)
+  })
+})


src/modules/Range.js
```diff
@@ -102,11 +102,21 @@ class Range {
...
         // Scale the Y axis to the min..max within the possibly zoomed X axis domain.
+        // Subpixel rounding in brush selection may produce a pixel→timestamp
+        // conversion that is slightly off exact data point timestamps, causing
+        // boundary points to be excluded from the Y range. (#5251)
+        let onePixelInDomain = 0
+        if (
+          gl.maxX !== gl.minX &&
+          this.w.layout.gridWidth > 0
+        ) {
+          onePixelInDomain = (gl.maxX - gl.minX) / this.w.layout.gridWidth
```
junedchhipa, line 113: This evaluates to `-Infinity` on every re-render and `0` on the first render, never an actual pixel width.

`gl.minX` / `gl.maxX` are still the reset sentinels (`+/-Number.MAX_VALUE`) here: `Range.init()` calls `setYRange()` before `setXRange()`, and `setXRange()` is the only writer of those globals, while `resetGlobals()` and `fastUpdate()` wipe them before every render. `gl.maxX - gl.minX` therefore overflows to `-Infinity`, and the `gl.maxX !== gl.minX` guard above does not catch it because the two sentinels genuinely differ.

Downstream, `cnf.xaxis.min - onePixelInDomain` becomes `+Infinity` and `cnf.xaxis.max + onePixelInDomain` becomes `-Infinity`, so both scan loops run unconditionally and the window collapses to one index. On a 12-point datetime area chart with `zoomX(first, last - 1)` the window goes from `0..11` to `11..11` and the y scale from `380..500` to `539..541`.

Suggest computing the ratio from `cnf.xaxis.max - cnf.xaxis.min`, which is the only trustworthy domain at this point in the pipeline, and guarding with `Number.isFinite(...)` plus a positive-width check.

src/modules/Range.js
```diff
@@ -102,11 +102,21 @@ class Range {
       let lastXIndex = series[i].length - 1
       if (autoScaleYaxis) {
         // Scale the Y axis to the min..max within the possibly zoomed X axis domain.
+        // Subpixel rounding in brush selection may produce a pixel→timestamp
+        // conversion that is slightly off exact data point timestamps, causing
+        // boundary points to be excluded from the Y range. (#5251)
+        let onePixelInDomain = 0
+        if (
+          gl.maxX !== gl.minX &&
+          this.w.layout.gridWidth > 0
```
junedchhipa, line 111: `w.layout.gridWidth` is stale at this point: `plotCoords()` runs after `coreCalculations()`. Even once the `-Infinity` overflow above is fixed, this ratio is `0` on the first render, and on a re-render it comes from the pre-resize layout whenever a resize and a zoom coincide.

tests/unit/brush-autoscaleyaxis-boundary.spec.js
```diff
@@ -0,0 +1,136 @@
...
+      },
+    })
+
+    // Simulate a brush selection that ends slightly before the last data point
+    chart.zoomX(data[0][0], brushMax)
+
+    // The Y-axis max should include the last point (540) even though
+    // brushMax is slightly before lastX, because the difference is
+    // less than one rendered pixel in the X domain.
+    expect(chart.w.globals.yAxisScale[0].niceMax).toBeGreaterThanOrEqual(lastY)
```
junedchhipa, line 81: This is the only assertion in the test, and the degenerate single-point window produced by the `-Infinity` bug satisfies it (`niceMax` lands at `541`). That is why CI is green despite the regression.

A min-side assertion would catch it, e.g. `expect(...niceMin).toBeLessThanOrEqual(388)` alongside the existing max check.

tests/unit/brush-autoscaleyaxis-boundary.spec.js
```diff
@@ -0,0 +1,136 @@
...
+        max: data[data.length - 1][0],
+      },
+    })
+
+    // Zoom to first 2 points only — the last point (500) is far outside
+    chart.zoomX(data[0][0], data[1][0])
+
+    // The Y-axis max should be based on the max of the first 2 points (200),
+    // not the last point (500) which is far outside the range.
+    expect(chart.w.globals.yAxisScale[0].niceMax).toBeLessThan(500)
```
junedchhipa, line 134: This test never exercises the new code path. There is no `chart.width` in these options, so `w.layout.gridWidth` is `NaN` under jsdom, the `> 0` guard short-circuits, and `onePixelInDomain` stays `0`. The spec passes byte-identically with `Range.js` reverted to `main`.

Worth setting an explicit width here (as the first test does) so the guard is actually reached.

junedchhipa (changes requested):
Thanks for digging into #5251. Unfortunately the patch as written does not work, and it actively regresses `autoScaleYaxis`. I reproduced this with instrumentation rather than only reading the diff.

### The core problem

`onePixelInDomain` never evaluates to a real pixel width. It is `-Infinity` on every re-render, and `0` on the first render.

`gl.minX` / `gl.maxX` are still the reset sentinels (`+/-Number.MAX_VALUE`) at the moment `getMinYMaxY()` runs, because `Range.init()` calls `setYRange()` before `setXRange()`, and `setXRange()` is the only writer of those globals. `resetGlobals()` (`Data.js`) and `fastUpdate()` (`apexcharts.js`) wipe them before every render. So:

- `gl.maxX - gl.minX` overflows to `-Infinity`
- the `gl.maxX !== gl.minX` guard does not catch it (the sentinels genuinely differ)
- `cnf.xaxis.min - onePixelInDomain` becomes `+Infinity` and `cnf.xaxis.max + onePixelInDomain` becomes `-Infinity`, so **both loop conditions are unconditionally true** and the scan window collapses to a single point

Measured on a 12-point datetime area chart with `zoomX(first, last - 1)`:

| | scan window | y scale |
|---|---|---|
| `main` | `0..11` | `380..500` |
| this branch | `11..11` | `539..541` |

Every point but one ends up off-chart.

### Why CI is green

Both new tests pass for the wrong reasons, so they do not protect the fix:

1. The boundary test only asserts the upper bound, which the degenerate single-point window happens to satisfy.
2. The "no false positive" test never executes the new code at all: without `chart.width`, `w.layout.gridWidth` is `NaN` under jsdom, the `> 0` guard short-circuits, and `onePixelInDomain` stays `0`. That test passes byte-identically with `Range.js` reverted to `main`.

### Suggested direction

At this point in the pipeline the only trustworthy domain is `cnf.xaxis.max - cnf.xaxis.min`, so derive `onePixelInDomain` from that instead of the globals. Please also gate it with `Number.isFinite(...)` plus a positive-width check rather than `gl.maxX !== gl.minX`, and note that `w.layout.gridWidth` is stale here (`plotCoords()` runs after `coreCalculations()`), so it is `0` on first render and reflects the pre-resize layout when a resize and a zoom coincide.

Once that is sorted, please strengthen both tests so they fail against `main`. That is the quickest way to confirm the fix is real.

junedchhipa:
this [commit](https://github.com/apexcharts/apexcharts.js/commit/26037e5710dee41cb786ce38749bad20957364f9) fixes #5251 
closing this PR in favor of that

PR #5261, opened 2026-08-08
fix: handle null values in area chart with line threshold colors (#5209)

### LINKED ISSUE #5209: Incorrect rendering when null value exists in the data with Threshold options
### Description

Incorrect rendering when `null` value exists in the data and colorAboveThreshold, colorBelowThreshold option is used.

### Steps to Reproduce

A chart with the following data and options will render incorrectly

```
      series: [
        {
          data: [60, 63, 60, 87, -50, -50, null, 100, 120, 138],
        },
      ],
      chartOptions: {
        chart: {
          type: 'area',
          height: 350,
        },
        plotOptions: {
          line: {
            isSlopeChart: false,
            colors: {
              threshold: 0,
              colorAboveThreshold: 'rgba(0, 136, 238, 0.30)',
              colorBelowThreshold: 'rgba(255, 0, 0, 0.30)',
            },
          },
        },
        dataLabels: {
          enabled: false,
        },
        xaxis: {
          categories: [
            'Korea',
            'Canada',
            'United Kingdom',
            'Netherlands',
            'Italy',
            'France',
            'Japan',
            'United States',
            'China',
            'Germany',
          ],
        },
      },
    };
```

### Screenshots

Image

### Reproduction Link

https://stackblitz.com/edit/vue-bar-chart-gmrjjet5?file=src%2FApp.vue

waterWang:
## Description

Fixes incorrect rendering when data contains `null` values and `colorAboveThreshold` / `colorBelowThreshold` options are used in an area chart.

### The Problem

When an area chart has a `null` in its data, the area path is split into multiple SVG `<path>` segments. Each segment references the same gradient fill, but because the gradient uses the default `objectBoundingBox` coordinate system, the color stop offset is applied relative to **each segment's own bounding box**. This causes the blue-to-red threshold transition to appear at **different absolute Y positions** across segments.

For example, with data `[60,63,60,87,-50,-50,null,100,120,138]` and threshold `0`:
- **Segment 1** (values -50 to 87): transition at ~73.4% of the segment's bounding box
- **Segment 2** (values 100 to 138, all above threshold): transition also at ~73.4% of its bounding box, which is a different absolute Y position, incorrectly showing red in the bottom portion of an all-positive segment

### The Fix

1. **`computeColorStops`** (Fill.js): Skip null/undefined values so they are not treated as `0` (JavaScript coercion: `null >= 0` is `true`).

2. **`drawGradient`** (Graphics.js): When the chart has null values AND is an area chart, use `gradientUnits="userSpaceOnUse"` with the grid height. This ensures the gradient stop offset maps consistently across **all path segments**, regardless of each segment's individual bounding box.

### Tests

- Added 3 regression tests in `tests/unit/issue-5209-null-threshold.spec.js`
- All 1979 unit tests pass (3 new + 1976 existing)

Fixes #5209

### DIFF
--- a/src/modules/Fill.js
+++ b/src/modules/Fill.js
@@ -123,6 +123,8 @@ class Fill {
     let minNegative = null
 
     for (const value of data) {
+      // skip null/undefined values to avoid treating null as 0 (null >= 0 is true in JS)
+      if (value === null || value === undefined) continue
       if (value >= multiColorConfig.threshold) {
         if (maxPositive === null || value > maxPositive) {
           maxPositive = value
--- a/src/modules/Fill.js
+++ b/src/modules/Fill.js
@@ -114,48 +114,56 @@ class Fill {
   }
 
   /**
-   * @param {number[]} data
+   * The y-axis window a series is plotted against. This is the space the
+   * threshold gradient is measured in, since that gradient spans the plot area.
+   *
+   * @param {number} realIndex
+   */
+  getSeriesAxisRange(realIndex) {
+    const w = this.w
+    const yaxisIndex = w.globals.seriesYAxisReverseMap?.[realIndex] ?? 0
+
+    return {
+      minY: Utils.isNumber(w.globals.minYArr[realIndex])
+        ? w.globals.minYArr[realIndex]
+        : w.globals.minY,
+      maxY: Utils.isNumber(w.globals.maxYArr[realIndex])
+        ? w.globals.maxYArr[realIndex]
+        : w.globals.maxY,
+      reversed: !!w.config.yaxis[yaxisIndex]?.reversed,
+    }
+  }
+
+  /**
+   * Builds the two stops that split a series into an above- and a
+   * below-threshold color.
+   *
+   * The stops are painted into a vertical gradient anchored to the plot area
+   * (see Graphics.drawGradient's verticalUserSpace branch), so the boundary is
+   * positioned over the *axis* range. Deriving it from the data range instead
+   * puts the transition on the wrong value whenever the axis extends past the
+   * data, which is the common case: an explicit yaxis.min/max, a nice scale, or
+   * an axis shared with another series.
+   *
+   * @param {{ minY: number, maxY: number, reversed?: boolean }} axisRange
    * @param {Record<string, any>} multiColorConfig
    */
-  computeColorStops(data, multiColorConfig) {
+  computeColorStops(axisRange, multiColorConfig) {
     const w = this.w
-    let maxPositive = null
-    let minNegative = null
+    const { threshold, colorAboveThreshold, colorBelowThreshold } =
+      multiColorConfig
+    const { minY, maxY } = axisRange
 
-    for (const value of data) {
-      // skip null/undefined values to avoid treating null as 0 (null >= 0 is true in JS)
-      if (value === null || value === undefined) continue
-      if (value >= multiColorConfig.threshold) {
-        if (maxPositive === null || value > maxPositive) {
-          maxPositive = value
-        }
-      } else {
-        if (minNegative === null || value < minNegative) {
-          minNegative = value
-        }
-      }
+    const span = maxY - minY
+
+    // A flat axis has no interior to split: the series sits wholly on one side
+    // of the threshold.
+    let offset = span === 0 ? (threshold > maxY ? 0 : 100) : ((maxY - threshold) / span) * 100
+
+    // A reversed axis puts the low values at the top, mirroring the boundary.
+    if (axisRange.reversed) {
+      offset = 100 - offset
     }
-
-    if (maxPositive === null) {
-      maxPositive = multiColorConfig.threshold
-    }
-    if (minNegative === null) {
-      minNegative = multiColorConfig.threshold
-    }
-
-    let totalRange =
-      maxPositive -
-      multiColorConfig.threshold +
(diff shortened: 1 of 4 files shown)


junedchhipa (commented):
Thanks for digging into #5209, and for the reproduction. The diagnosis is right: when nulls split an area/line into several path elements, an `objectBoundingBox` gradient is resolved against *each segment's own* bbox, so the threshold boundary lands at a different data value in every segment. Switching to `userSpaceOnUse` is the correct direction.

I went through the branch and verified a few things in jsdom before merging, and there are three problems that need fixing. Since they touch the central offset mapping rather than being local tweaks, I'm going to push the follow-up commits to this branch myself rather than send you round another review cycle. Your commit stays in the history. Writing up what I found so the reasoning is on the record:

**1. The gate is broader than the bug.** `w.globals.hasNullValues && w.config.chart.type === 'area'` re-anchors *every* vertical gradient in the chart, not just the threshold one. A plain `fill: { type: 'gradient' }` area with a null and no threshold colors configured now gets `gradientUnits="userSpaceOnUse" y2="288.82"` instead of `y2="1"`. Same for a `bar` series inside an area-typed combo chart, and for series that contain no nulls at all, since `hasNullValues` is chart-global rather than per-series. `samples/source/area/area-with-missing-data.xml` (gradient + nulls) has an e2e snapshot that shifts because of this.

**2. The threshold moves.** This is the important one. `computeColorStops` derives the offset from the **data** range:

```
offset = (maxPositive - threshold) / (maxPositive - minNegative)
```

Under `objectBoundingBox` that was exact, because the path's bbox *is* the data range. But the diff maps that same offset over the **axis** range (`y2 = gridHeight`), and those two ranges are not the same thing. With the data in your own test the color boundary lands at y 212.0 while true zero is at y 216.6; add `yaxis: { min: -500, max: 500 }` and it is off by 67.6px. The fix trades a per-segment inconsistency for a consistent-but-wrong boundary.

Once the gradient is anchored to the plot area, the offset has to be expressed in that same space:

```
offset = (maxY - threshold) / (maxY - minY)     // axis range, not data range
```

which also means `computeColorStops` no longer needs to scan the data at all.

**3. The `Fill.js` guard doesn't cover the real input.** `value === null || value === undefined` misses `NaN`, which is what non-numeric and undefined data actually becomes by the time it reaches `w.seriesData.series`. `data: [60, undefined, 30]` with `threshold: 50` still emits `<stop offset="NaN">`. `value == null || !Number.isFinite(value)` covers it.

Two smaller notes:

- All three new tests still pass with the `Fill.js` hunk reverted. For the `null` case that hunk is actually a no-op: assigning `null` to `maxPositive`/`minNegative` leaves them `=== null`, so the next value reassigns unconditionally either way. Worth knowing when a test is pinning the behaviour you think it is.
- The same defect exists on `chart.type: 'line'`, which is the primary consumer of `plotOptions.line.colors`: two stroke paths sharing one `objectBoundingBox` gradient. The `chart.type === 'area'` gate leaves that case broken. I'll fold the line case into this PR.

Plan for the follow-up commits: drop the `hasNullValues` gate and key the user-space anchoring off the threshold feature itself (so it is deterministic and does not touch unrelated gradients), recompute the offset over the axis range, extend to line, tighten the guard, and add tests that actually fail without the fix.

Appreciate you taking the first pass at this.

junedchhipa:
Pushed the follow-up to this branch (229660979). What changed on top of your commit:

- `computeColorStops` now takes the y-axis window instead of the series data and positions the boundary at `(maxY - threshold) / (maxY - minY)`, matching the space the plot-area-anchored gradient actually spans. Verified against an independent oracle: a y-axis annotation placed at the threshold. With the data from your test the transition now lands at y 216.615, exactly where the annotation sits, and it stays there under `yaxis: { min: -500, max: 500 }`.
- User-space anchoring is keyed off the threshold feature itself (threaded as `verticalUserSpace` through `handleGradientFill` into `drawGradient`) rather than off `hasNullValues`, so it no longer depends on whether the chart happens to contain a null and no longer touches unrelated gradients.
- Removed the `chart.type === 'area'` gate, so `chart.type: 'line'` is covered.
- Reversed axes mirror both the boundary and the stop order.
- The two stops are emitted in ascending offset order instead of relying on the SVG rule that clamps an out-of-order offset up to its predecessor.
- Since `computeColorStops` no longer reads the series, non-finite data cannot reach the offset math at all, so the `Fill.js` guard came out with the data scan.

One correction to my review above: I said `samples/source/area/area-with-missing-data.xml` had an e2e snapshot that would shift. That was wrong. The default `fill.gradient.type` is `horizontal`, so that sample never enters the vertical branch, and its snapshot passes on your commit unchanged. The underlying point still holds (I confirmed the over-broad gate by setting `gradient.type: 'vertical'` explicitly), but the sample I pointed at was not affected. That also means the test I wrote for it initially passed for the wrong reason, which is now fixed with an explicit `gradient.type`.

Checks: unit suite 2148 passing, lint clean, `tsc --noEmit` output byte-identical to main's baseline, and the `area-with-negative` / `line-with-negative` / `area-with-missing-data` e2e snapshots pass unchanged. The 9 tests in `issue-5209-null-threshold.spec.js` were each confirmed to fail without the fix (8 of 9 fail against your commit, 6 of 8 against pre-PR main).

Thanks again for finding and diagnosing this one.

junedchhipa (approved):

PR #5262, opened 2026-08-08
fix: correct left-edge guard for dataLabels textAnchor:start in horizontal bar charts (Closes #5094)

### LINKED ISSUE #5094: Datalabel has misaligned position when datalabel text is longer than bar width in horizontal bar chart
### Description

Data label is positioned on the x-axis to 0 when the text width is greater than the bar width and the text anchor is set to "start". This causes the data label to be misaligned with the rest of the data labels in the chart.

I believe this part of the code is causing this issue:
```javascript
        if (dataLabelsX - textRects.width < 0) {
          dataLabelsX = valIsNegative
            ? textRects.width + strokeWidth
            : strokeWidth
        }
```
https://github.com/apexcharts/apexcharts.js/blob/main/src/charts/common/bar/DataLabels.js#L502-L505

### Steps to Reproduce

1. Create a horizontal bar chart with data with more than 4 decimal places and with a large difference between the largest and smallest data point.

Example:
```javascript
var options = {
  chart: {
    width: "100%",
    type: "bar",
    parentHeightOffset: 0,
    toolbar: {
      show: false
    }
  },
  series: [
    {
      name: "categories",
      data: [17332, 2235, 1251, 1093]
    }
  ],
  plotOptions: {
    bar: {
      horizontal: true,
      distributed: true,
      barHeight: "80%",
      dataLabels: {
        position: "top",
        hideOverflowingLabels: false
      }
    }
  },
  xaxis: {
    labels: { show: false },
    axisBorder: { show: true, color: "#D5D2D5" },
    axisTicks: { show: false },
    crosshairs: { show: false },
    categories: ["roupas", "calcados", "bolsas", "acessorios"]
  },
  yaxis: {
    labels: {
      minWidth: 80,
      style: { fontSize: "14px", fontWeight: 500 }
    },
    axisTicks: { show: false },
    crosshairs: { show: false }
  },
  dataLabels: {
    textAnchor: "start",
    offsetX: 10,
    style: {
      colors: ["#222222"],
      fontSize: "14px",
      fontWeight: 700
    }
  },
  legend: { show: false },
  grid: { show: false, padding: { top: 0, bottom: 0 } },
  tooltip: { enabled: false },
  colors: ["#06D4BB", "#075E6F", "#8C1D47", "#F0967C"]
};

chart.render();
```

### Expected Behavior
All data labels must align at the x position (bar width) at the end of the respective bar + offsetX + stroke width

### Actual Behavior
The x position of the data label that has the text width greater than the bar width is being set to the stroke width value

### Screenshots

Image

### Reproduction Link

https://codepen.io/Gabi-chan/pen/KwdxdLY

waterWang:
## Summary

Fixes #5094 — Data label is misaligned when the label text is wider than the bar in a horizontal bar chart with `textAnchor: 'start'`.

### Root cause

In `src/charts/common/bar/DataLabels.js`, the left-edge overflow guard for `textAnchor: 'start'` used:

```js
if (dataLabelsX - textRects.width < 0) { ... }
```

This condition is correct for `textAnchor: 'end'` (where text extends **left** from the anchor), but **wrong** for `textAnchor: 'start'`, where the text extends **right** from `dataLabelsX`.

As a result, whenever a label's text was wider than the bar tip position (`dataLabelsX`), the condition fired even though the anchor was comfortably inside the plot — pinning the label to `strokeWidth + offX` (near the left edge) and misaligning it with the other labels in the chart.

### Fix

Change the guard to check the anchor point itself:

```js
if (dataLabelsX < 0) { ... }
```

Now the left-edge guard only fires when the anchor point is genuinely off-screen to the left — which is the correct behavior for `textAnchor: 'start'`.

## Tests

Added `tests/unit/issue-5094-datalabel-horizontal-bar.spec.js` with 4 regression tests verifying:
- All 4 data labels render
- All labels have positive x positions inside the plot
- Label x positions are ordered proportionally to their values
- Labels are not pinned to the left edge

All 1980 unit tests pass (92 files, 0 new failures).

## Type of change

### DIFF
--- a/src/charts/common/bar/DataLabels.js
+++ b/src/charts/common/bar/DataLabels.js
@@ -560,12 +560,13 @@ export default class BarDataLabels {
 
     if (!w.config.chart.stacked) {
       if (dataLabelsConfig.textAnchor === 'start') {
-        if (dataLabelsX - textRects.width < 0) {
-          // Pin the label to the plot edge, but keep honoring the user's
-          // offsetX from that edge — otherwise a `position:'bottom'` +
-          // `textAnchor:'start'` label (whose dataLabelsX equals the offset for
-          // left-rooted bars) silently loses any offsetX smaller than its own
-          // text width. offX is 0 by default, so this is a no-op unless set.
+        if (dataLabelsX < 0) {
+          // Pin the label to the plot edge when the anchor point itself
+          // is off-screen to the left. textAnchor:'start' means the text
+          // extends to the RIGHT from dataLabelsX, so only the anchor
+          // position matters — not the text width. Previously used
+          // `dataLabelsX - textRects.width < 0` which incorrectly pinned
+          // labels whose text was wider than the bar (issue #5094).
           dataLabelsX = valIsNegative
             ? textRects.width + strokeWidth - offX
             : strokeWidth + offX
--- a/tests/unit/issue-5094-datalabel-horizontal-bar.spec.js
+++ b/tests/unit/issue-5094-datalabel-horizontal-bar.spec.js
@@ -1,0 +1,88 @@
+import { createChartWithOptions } from './utils/utils.js'
+
+// Integration test for #5094 — Datalabel misaligned when text is wider than bar
+// in horizontal bar chart with textAnchor:'start'.
+//
+// In jsdom, getTextRects returns a fixed width of ~10px, so the original bug
+// condition (dataLabelsX - textRects.width < 0) cannot be triggered. However,
+// the fix is semantically correct: for textAnchor 'start' the text extends
+// RIGHT from the anchor, so the left-edge overflow condition must check
+// `dataLabelsX < 0` (the anchor itself), not `dataLabelsX - textRects.width < 0`
+// (which is designed for textAnchor 'end').
+//
+// This test verifies that:
+// 1. All data labels are rendered (4 bars → 4 labels)
+// 2. All x positions are positive and inside the plot
+// 3. x positions are monotonic with their values (proportional)
+
+describe('datalabel position in horizontal bar (issue #5094)', () => {
+  function horizontalBarChart(overrides = {}) {
+    return createChartWithOptions({
+      chart: {
+        type: 'bar',
+        width: '800px',
+        height: 300,
+        parentHeightOffset: 0,
+        toolbar: { show: false },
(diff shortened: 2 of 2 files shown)


junedchhipa (changes requested):
Thanks for digging into #5094 — the direction here is right for positive values, but the change over-applies and the new test doesn't currently guard the fix. Two things before this can land.

### 1. Negative bars get pushed outside the plot

In `src/charts/common/bar/DataLabels.js`, dropping `- textRects.width` from the left-edge guard is correct for positive values (anchor stays `start`, text extends right, so only the anchor x matters). But the same branch also handles negatives, and `drawCalculatedDataLabels` flips `start` -> `end` when `val < 0` on horizontal bars. With anchor `end` the label extends *left* from `dataLabelsX`, so the width term was load-bearing there. The branch's own negative pin (`textRects.width + strokeWidth - offX`) and the right-edge sibling (`gridWidth - strokeWidth`) both confirm the anchor-`end` semantics.

Verified in Chromium against `samples/source/tests/position-top-text-anchor-start-series-neg.xml`: the `India: -1380` label moves from x=63.39 (bbox left -2.6) on `main` to x=49.04 (bbox left -16.99) with this PR, i.e. ~17px now sits outside the plot area. A tighter case (`[-17332, -16000, -15000]` with `xaxis: { min: -17332, max: 0 }`) pushes the label fully off-screen at x=0, where `main` pinned it at x=72.

Suggested fix — keep the width term only on the anchor-`end` path:

```js
if (valIsNegative ? dataLabelsX - textRects.width < 0 : dataLabelsX < 0) {
```

The e2e snapshot matrix doesn't catch this because the per-sample tolerance is 5% of pixels.

### 2. The new spec passes without the fix

`tests/unit/issue-5094-datalabel-horizontal-bar.spec.js` is green (4/4) against the pre-fix source — I reverted `DataLabels.js` and re-ran to confirm. The reason is noted in the spec's own header: the jsdom `getBBox` stub in `tests/unit/setup.js` returns a fixed 10px width, so the `textRects.width` term can never influence the outcome.

Overriding `SVGGraphicsElement.prototype.getBBox` with a length-proportional width makes both the original bug and the negative-value regression above reproducible in unit tests. Please also add coverage for negative series values.

Minor: the new test file is missing a trailing newline.

PR #5263, opened 2026-08-09
fix: honor xaxis.stepSize when xaxis.max is also set (Closes #5082)

### LINKED ISSUE #5082: Xaxis "max" value is ignored when also providing a "stepSize"
### Description

Create a bubble chart with a XAxis which sets both the "max" value and "stepSize". E.g. with this setup:

<Details>
<Summary> JS Code </Summary>

```
function generateData(basetime, count, yrange) {
  var i = 0
  var series = []

  while (i < count) {
    var x = Math.floor(Math.random() * (750 - 1 + 1)) + 1
    var y =
      Math.floor(Math.random() * (yrange.max - yrange.min + 1)) + yrange.min
    var z = Math.floor(Math.random() * (75 - 15 + 1)) + 15 // bubble size

    series.push([x, y, z])
    i++
  }

  return series
}

var options = {
  series: [
    {
      name: 'Bubble1',
      data: generateData(new Date('11 Feb 2017 GMT').getTime(), 20, {
        min: 10,
        max: 60,
      }),
    },
    {
      name: 'Bubble2',
      data: generateData(new Date('11 Feb 2017 GMT').getTime(), 20, {
        min: 10,
        max: 60,
      }),
    },
    {
      name: 'Bubble3',
      data: generateData(new Date('11 Feb 2017 GMT').getTime(), 20, {
        min: 10,
        max: 60,
      }),
    },
    {
      name: 'Bubble4',
      data: generateData(new Date('11 Feb 2017 GMT').getTime(), 20, {
        min: 10,
        max: 60,
      }),
    },
  ],
  chart: {
    height: 350,
    type: 'bubble',
  },
  dataLabels: {
    enabled: false,
  },
  fill: {
    opacity: 0.8,
  },
  title: {
    text: 'Simple Bubble Chart',
  },
  xaxis: {
    max: 200,
    stepSize: 50,
    min: 0,
    type: 'double',
  },
  yaxis: {
    max: 70,
  },
}

var chart = new ApexCharts(document.querySelector('#chart'), options)
chart.render()

```

</Details>

Possibly related issue i found: https://github.com/apexcharts/apexcharts.js/issues/4670

### Steps to Reproduce

1. Create the buble chart with some example data
2. Set axis "max" and "stepSize" properties (e.g. 50 and 200)
3. See that max value is ignored in that case

### Expected Behavior

The "max" value should still apply even if the "stepSize" is set as well.

### Actual Behavior

The "max" value is completly ignored with the "stepSize" options set.

### Screenshots

This chart has "max" on XAxis set to 200:

Image

### Reproduction Link

(https://codepen.io/neurotoxin41241/pen/raOegBL

waterWang:
## Summary

Fixes #5082 — **Xaxis "max" value is ignored when also providing a "stepSize"**.

When a user sets both `xaxis.max` and `xaxis.stepSize` on a numeric (or bubble/double) chart, the `stepSize` was being silently dropped. The root cause was in `Scales.setXScale()`:

```js
// before
w.config.xaxis.max === undefined ? w.config.xaxis.stepSize : undefined,
```

The ternary discarded `stepSize` whenever `xaxis.max` was defined, so tick steps fell back to `range / ticks` instead of honoring the user's explicit step size. Both options are independent and should be respected together — `max` pins the upper bound while `stepSize` controls the tick interval.

## Change

```js
// after
w.config.xaxis.stepSize,
```

`linearScale()` already handles `step === undefined` (falls back to `range / ticksNum`), so passing `stepSize` unconditionally is safe and preserves the existing behavior for charts that don't set it.

## Tests

- All existing unit tests pass (1980 passed).
- The 3 failing suites (`license-gating`, `license-manager`, `unit-chart`) are pre-existing environment failures (`No such built-in module: node:`) and fail identically on `main` without this change.

## Verification

Given options `xaxis: { min: 0, max: 200, stepSize: 50, type: 'double' }` on a bubble chart, the x-axis now renders ticks at 0, 50, 100, 150, 200 — honoring both the `max` bound and the explicit `stepSize`.

Closes #5082

### DIFF
--- a/src/modules/Scales.js
+++ b/src/modules/Scales.js
@@ -647,7 +647,7 @@ export default class Scales {
         maxX,
         ticks,
         0,
-        w.config.xaxis.max === undefined ? w.config.xaxis.stepSize : undefined,
+        w.config.xaxis.stepSize,
       )
     }
     return gl.xAxisScale


junedchhipa:
I have to request changes: this line is a deliberate guard, and removing it re-introduces the exact behaviour reported in #5082.

## Why the change breaks `max`

`linearScale()` does not clamp to `yMax`. It emits `ticks + 1` values starting at `minX` and stepping by `step`:

```js
let v = yMin
while (tickCount >= 0) {
  result.push(v)
  v = Utils.preciseAddition(v, step)
  tickCount -= 1
}
```

`ticks` is `gl.xTickAmount`, derived from chart width / dataPoints / `tickAmount` in `Range.js`, and it has no relationship to `stepSize`. So as soon as `stepSize` is passed through unconditionally, `niceMax` becomes `minX + xTickAmount * stepSize`, which almost never equals the user's `max`.

Running the config from your PR description (bubble, `type: 'double'`, `min: 0, max: 200, stepSize: 50`):

| build | `globals.xAxisScale.result` |
| --- | --- |
| `main` | `[0, 20, 40, 60, 80, 100, 120, 140, 160, 180, 200]` (max honoured, stepSize ignored) |
| this PR | `[0, 50, 100, 150, 200, 250, 300, 350, 400, 450, 500]` (max overshot by 2.5x) |

With the patch the rendered x-axis labels read `0 ... 500` across a grid whose data range is `0 ... 200`, so every label is misaligned with the data underneath it. A second case, `min: 0, max: 100, stepSize: 30` with 12 points: `main` gives `0 ... 99.99`, the PR gives `0, 30, 60, 90, 120, ... 330`.

The `0, 50, 100, 150, 200` result quoted in the PR only appears when `xTickAmount` happens to equal `range / stepSize` (for example a numeric axis with exactly 5 data points, where `ticks = dataPoints - 1`). It does not generalise.

## This reverts the fix for the issue it cites

The ternary was added intentionally in `ba757dba5` (2025-07-25, shipped in v5.10.0), two days after #5082 was filed on 2025-07-23. The issue's complaint is that **`max` is ignored** ("See that max value is ignored in that case", plus a screenshot of the axis running well past 200). This PR makes `stepSize` win and `max` lose, which is the reported bug rather than the fix. On current `main` the reporter's exact config already ends the axis at `max`.

## What a correct fix would need

Honouring both options means deriving the tick count from the range and the step, and keeping `gl.xTickAmount` in sync so the grid lines match the labels. Roughly:

```js
const { stepSize, max } = w.config.xaxis
let ticks = gl.xTickAmount
if (stepSize && max !== undefined && w.config.xaxis.tickAmount === undefined) {
  ticks = Math.ceil(Math.abs(maxX - minX) / stepSize)
  gl.xTickAmount = ticks
}
gl.xAxisScale = this.linearScale(minX, maxX, ticks, 0, stepSize)
```

Two design questions need answering before any patch lands:

1. Non-divisible ranges (`max: 100, stepSize: 30`): should the last tick overshoot to 120, be clamped to 100, or should `stepSize` be GCD-adjusted? `niceScale()` already solves this for the y-axis (`Utils.getGCD` in `Scales.js`), so the x path should mirror that rather than invent a third behaviour.
2. Precedence when `tickAmount`, `stepSize` and `max` are all set, since all three cannot hold simultaneously.

## Other notes

* **No test.** `tests/unit/xaxis.spec.js` and `tests/unit/scales.spec.js` are the natural homes. All 93 existing tests there pass with the regression in place, so "all existing unit tests pass" is not evidence of correctness here. A fix should land with a case asserting `xAxisScale.result` for the issue config.
* The reasoning in the description ("`linearScale()` already handles `step === undefined`") is accurate but addresses the wrong failure mode. The problem is not the undefined branch, it is the unbounded tick loop.
* Unrelated but visible while reproducing: on `main` the issue config produces a final label of `200.07` instead of `200`, from the step-rounding block in `linearScale()`. Pre-existing, worth a separate issue.

Happy to review a follow-up that derives the tick count from `stepSize` and adds a regression test.

PR #5264, opened 2026-08-09
fix: clamp column data labels X to prevent covering y-axis labels (Closes #5056)

### LINKED ISSUE #5056: Data label covering part of a value on the y-axis
### Description
When having multiple series and data labels, sometimes the data labels covers a part of the value on the y-axis.

### Steps to Reproduce

1. Create a graph with categories such as `categories: ["v21", "v22", "v23"]`
2. ```
Add series such as: series: [
    {
      name: "x",
      data: [859, 813, 795] 
    },
    {
      name: "y",
      data: [5749, 5100, 4453]
    },
    {
      name: "z",
      data: [7059, 6739, 6629]
    },
    {
      name: "a",
      data: [15166, 14481, 13523]
    }
  ],
```
3. Enable data labels

### Expected Behavior
My expectatiions is that the values on the y axis are all visible and easy to read
### Actual Behavior
A data label is covering a part of the value on the y axis so I can't see the value.
### Screenshots

![Image](https://github.com/user-attachments/assets/68a4362c-495a-4819-b90f-f74944177f18)

### Reproduction Link
https://codepen.io/martybof/pen/myJpXKW

### PR DESCRIPTION
**Pull request by waterWang**
Closes #5056

## Problem
When a vertical (column) bar chart has multiple series with data labels enabled, the data label for the leftmost bar can extend past the plot area's left edge and cover the y-axis labels. The rightmost label can similarly spill off the right edge.

## Root Cause
`calculateColumnsDataLabelsPosition` in `src/charts/common/bar/DataLabels.js` clamps the **Y** position (lines 382-388) but does **not** clamp the **X** position. The horizontal-bar counterpart `calculateBarsDataLabelsPosition` already clamps X (lines 562-590), so vertical columns were missing this protection.

## Fix
Added `textAnchor`-aware X-edge clamping to `calculateColumnsDataLabelsPosition`, mirroring the existing horizontal-bar logic:
- `middle` → keep `dataLabelsX ± width/2` inside `[0, gridWidth]`
- `start` → keep `dataLabelsX` and `dataLabelsX + width` inside the grid
- `end` → keep the label inside the grid

## Testing
- Existing unit suite passes (1976 tests).
- The 3 failing suites (`license-*`, `unit-chart`) are pre-existing Node built-in module failures unrelated to this change.

### DIFF
--- a/src/charts/common/bar/DataLabels.js
+++ b/src/charts/common/bar/DataLabels.js
@@ -387,6 +387,30 @@ export default class BarDataLabels {
       }
     }
 
+    // Clamp dataLabelsX to prevent labels from extending past the grid
+    // edges and covering the y-axis labels or spilling off the right side.
+    // This mirrors the left-edge/right-edge clipping already present in
+    // calculateBarsDataLabelsPosition (horizontal bars) at lines 562-590.
+    if (dataLabelsConfig.textAnchor === 'middle') {
+      if (dataLabelsX - textRects.width / 2 < 0) {
+        dataLabelsX = textRects.width / 2 + strokeWidth
+      } else if (dataLabelsX + textRects.width / 2 > w.layout.gridWidth) {
+        dataLabelsX = w.layout.gridWidth - textRects.width / 2 - strokeWidth
+      }
+    } else if (dataLabelsConfig.textAnchor === 'start') {
+      if (dataLabelsX < 0) {
+        dataLabelsX = strokeWidth
+      } else if (dataLabelsX + textRects.width > w.layout.gridWidth) {
+        dataLabelsX = w.layout.gridWidth - textRects.width - strokeWidth
+      }
+    } else if (dataLabelsConfig.textAnchor === 'end') {
+      if (dataLabelsX < 1) {
+        dataLabelsX = textRects.width + strokeWidth
+      } else if (dataLabelsX + 1 > w.layout.gridWidth) {
+        dataLabelsX = w.layout.gridWidth - textRects.width - strokeWidth
+      }
+    }
+
     return {
       bcx,
       bcy: y,

### CODE REVIEW

**Comment by junedchhipa** [comment]
Thanks for the fix. The core idea is right and it does solve the reported case, but I checked out the branch and measured the rendered label `x` against the parent commit, and there are a few regressions / gaps I'd like sorted before merge.

Setup: grouped column chart, 400x300, `gridWidth: 363`, text measured at 7px/char.

| case | parent | this PR | |
|---|---|---|---|
| default (`textAnchor: 'middle'`) | first label spans `[-6.2, 84.8]` | `[0, 91]` | ✅ fix works |
| `orientation: 'vertical'` | 64.5 / 185.5 / 306.5 | 77 / 185.5 / **286** | ❌ labels shifted off their bars |
| `chart.stacked: true` | 60.5 / 181.5 / 302.5 | 77 / 181.5 / **286** | ❌ clamps where the Y clamp deliberately does not |
| `textAnchor: 'end'` | label spans `[-51.7, 39.3]` | unchanged | ❌ overflow not fixed, branch never fires |
| label wider than grid | 55.7 / 167.3 | **126 / 97** | ❌ label order inverts |

### 1. Breaks `plotOptions.bar.dataLabels.orientation: 'vertical'`

Vertical-orientation labels are drawn with `rotate(-90, x, y)`, so their horizontal footprint is `textRects.height` (~12px), not `textRects.width` (154px in my probe). The new clamp measures the unrotated width, so labels that were entirely inside the grid get shoved: the first moved +12.5px and the last -20.5px away from their bar centers. With narrower bars or longer text the label leaves its bar completely.

The same function already accounts for this a few lines above (`dataLabelsX + textRects.height / 2` in the `if (vertical)` block), so the clamp needs the same treatment:

```js
const labelWidth = vertical ? textRects.height : textRects.width
```

and then use `labelWidth` throughout the new block.

### 2. Missing the `!w.config.chart.stacked` guard

The Y clamp ~10 lines above is wrapped in `if (!w.config.chart.stacked)`, and so is the horizontal-bar block this PR mirrors. The new block isn't, so stacked columns now get X-shifted labels (60.5 → 77) sitting off-center inside their segment while the bar stays put. Either add the guard for consistency or call out explicitly why stacked should behave differently here.

### 3. The `textAnchor: 'end'` branch is dead code, and leaves the real overflow unfixed

With `end`, the text spans `[x - width, x]`. The copied conditions test the anchor point itself:

```js
if (dataLabelsX < 1) { ... } else if (dataLabelsX + 1 > w.layout.gridWidth) { ... }
```

For a column, `dataLabelsX` is a bar center and is essentially always inside `[1, gridWidth - 1]`, so neither branch ever fires. In my probe a `textAnchor: 'end'` label spanning `[-51.7, 39.3]` (well into the y-axis labels) passed through completely untouched. The correct tests are:

```js
if (dataLabelsX - textRects.width < 0) {
  dataLabelsX = textRects.width + strokeWidth
} else if (dataLabelsX > w.layout.gridWidth) {
  dataLabelsX = w.layout.gridWidth - strokeWidth
}
```

Note the existing horizontal-bar branch has the same bug, so copying it verbatim propagates it instead of mirroring working behavior.

### 4. Labels invert order when the label is wider than the grid

Because the branches are `else if`, only the left clamp applies and it lands right of where the right clamp would have put it. Two labels went from 55.7 / 167.3 to 126 / **97**, i.e. the second bar's label now renders to the *left* of the first bar's. Either skip the whole block when `textRects.width >= w.layout.gridWidth`, or clamp with a single `Math.min(Math.max(...))` so both bounds are respected.

### Smaller notes

- **`offsetX` is silently dropped at the edges.** With `dataLabels.offsetX: 20` the last label went 343.7 → 317.5. The horizontal implementation has explicit code (and a comment) to keep honoring `offX` from the pinned edge; worth matching.
- **Grouped labels can collapse onto the same x.** With `textAnchor: 'start'`, series A's and series B's third labels both clamped to exactly 272 and now draw on top of each other. Inherent to edge clamping, but another argument for scoping the clamp narrowly.
- **Stacked total labels (`totalDataLabelsX`) are still unclamped**, which is the same class of bug and arguably part of this issue's surface.
- **Please drop the hardcoded line numbers from the comment** ("at lines 562-590") - they rot on the next edit. Better still: extract the shared clamp into one helper used by both `calculateColumnsDataLabelsPosition` and `calculateBarsDataLabelsPosition`. The two copies are ~90% identical and already drifting (the `start` branch differs between them).
- **Please add a test.** `getBBox` is mocked to a constant 10x10 in `tests/unit/setup.js`, so a regression test needs a local `getBBox` override returning a length-proportional width (that's how I measured above). Good cases for `tests/unit/bar-chart.spec.js`: default anchor clamps to the grid edge, `orientation: 'vertical'` does *not* move, stacked stays unchanged.

For what it's worth, `bar-chart`, `datalabel-transition` and `dimensions` all pass on this head (88 tests), and the suites you mention are the known Node-version issue, unrelated to this change.

PR #5266, opened 2026-08-09
fix: compensate tooltip position for CSS zoom on the chart container (Closes #5043)

### LINKED ISSUE #5043: Bad tooltips positioning when container has the CSS zoom property != 100%.
### Description

When showing a chart in a container that has the `zoom` property set to something different than 100%, the tooltips on the line chart do not follow the mouse pointer properly. I suspect this is a problem that also occurs on other features that rely on the pointer feature. I reproduced this behavior on Line, Area, Pie charts but there could be more.

### Steps to Reproduce

1. Render a chart in some div
2. Apply zoom to 50% to said chart's parent
3. Move the mouse over the line chart

### Expected Behavior
The tooltip is shown on the point closest to the mouse pointer.

### Actual Behavior
The tooltip position is skewed to the left when zoom is less than 100%, to the right when zoom is more than 100%.

### Screenshots
Image

### Reproduction Link
https://codepen.io/meronz/pen/ByyXjrj

waterWang:
## Description

When the chart container has a CSS `zoom` property set to a value other than 1 (e.g. `zoom: 0.5`), the tooltip position is skewed — to the left when zoom < 100%, to the right when zoom > 100%. This happens because `getBoundingClientRect()` returns viewport-scaled values, but `style.left`/`style.top` on the tooltip (inside the zoomed container) are in CSS pixels which are further scaled by the zoom factor.

## Fix

Divide the tooltip position by the CSS zoom factor (`element.currentCSSZoom`) wherever the position is computed from viewport coordinates (`clientX`/`clientY` - `getBoundingClientRect()`). The zoom factor is obtained from the chart's `elWrap` element via the standard `currentCSSZoom` property.

## Changes

- **Position.js**: `computeTooltipPosition()` followCursor path and `moveStickyTooltipOverBars()`
- **Intersect.js**: `handleHeatTreeTooltip()`, `handleMarkerTooltip()`, and `handleBarTooltip()` followCursor paths
- **Tooltip.js**: `nonAxisChartsTooltips()` followCursor path
- **tests/unit/tooltip-positioning.spec.js**: 2 new tests for the CSS zoom compensation behavior

## Testing

All 1978 existing tests pass. New tests verify:
1. FollowCursor position is correctly divided by the zoom factor when `currentCSSZoom` is set
2. Fallback to zoom=1 when `currentCSSZoom` is undefined (backward compatibility)

Closes #5043

### DIFF
--- a/src/modules/tooltip/Intersect.js
+++ b/src/modules/tooltip/Intersect.js
@@ -184,13 +184,12 @@ class Intersect {
     }
     if (ttCtx.w.config.tooltip.followCursor) {
       const seriesBound = w.dom.elWrap.getBoundingClientRect()
+      const zoom = w.dom.elWrap.currentCSSZoom || 1
       x =
-        (w.interact.clientX ?? 0) -
-        seriesBound.left -
+        ((w.interact.clientX ?? 0) - seriesBound.left) / zoom -
         (x > w.layout.gridWidth / 2 ? ttCtx.tooltipRect.ttWidth : 0)
       y =
-        (w.interact.clientY ?? 0) -
-        seriesBound.top -
+        ((w.interact.clientY ?? 0) - seriesBound.top) / zoom -
         (y > w.layout.gridHeight / 2 ? ttCtx.tooltipRect.ttHeight : 0)
     }
 
@@ -266,7 +265,8 @@ class Intersect {
         const elGrid = ttCtx.getElGrid()
         if (!elGrid) return { x, y }
         const seriesBound = elGrid.getBoundingClientRect()
-        y = ttCtx.e.clientY + w.layout.translateY - seriesBound.top
+        const zoom = w.dom.elWrap.currentCSSZoom || 1
+        y = (ttCtx.e.clientY + w.layout.translateY - seriesBound.top) / zoom
       }
 
       ttCtx.marker.enlargeCurrentPoint(j, opt.paths, x, y)
@@ -632,11 +632,13 @@ class Intersect {
 
       if (w.config.tooltip.followCursor) {
         if (w.globals.isBarHorizontal) {
-          x = clientX - seriesBound.left + 15
+          const zoom = w.dom.elWrap.currentCSSZoom || 1
+          x = (clientX - seriesBound.left + 15) / zoom
           y = handleYForBars()
         } else {
+          const zoom = w.dom.elWrap.currentCSSZoom || 1
           x = handleXForColumns(x)
-          y = e.clientY - seriesBound.top - ttCtx.tooltipRect.ttHeight / 2 - 15
+          y = (e.clientY - seriesBound.top - ttCtx.tooltipRect.ttHeight / 2 - 15) / zoom
         }
       } else {
         if (w.globals.isBarHorizontal) {
--- a/src/modules/tooltip/Position.js
+++ b/src/modules/tooltip/Position.js
@@ -281,14 +281,21 @@ export default class Position {
       if (!elGrid) return null
       const seriesBound = elGrid.getBoundingClientRect()
 
-      x = ttCtx.e.clientX - seriesBound.left
+      // CSS zoom compensation: when the chart container has a CSS zoom
+      // property != 1, getBoundingClientRect returns viewport-scaled values
+      // but style.left/top are applied in the container's CSS pixel space
+      // which is further scaled by the zoom factor.  Divide by zoom so the
+      // tooltip follows the cursor at the correct position.
+      const zoom = w.dom.elWrap.currentCSSZoom || 1
+
+      x = (ttCtx.e.clientX - seriesBound.left) / zoom
       if (x > w.layout.gridWidth / 2) {
         x = x - ttW
         placement = 'left'
       } else {
         placement = 'right'
       }
-      y = ttCtx.e.clientY + w.layout.translateY - seriesBound.top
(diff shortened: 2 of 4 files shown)


junedchhipa:
Thanks for digging into this, and for the clear write-up. The diagnosis is right and `currentCSSZoom` is the correct API to reach for. I checked the branch out and ran the suite (green), but I think the arithmetic is inconsistent across the six call sites, and a few positioning bugs survive at `zoom != 1`.

### Ground truth

I probed Chromium directly with `zoom: 0.5` on a wrapper containing a 200x60 child:

```
wrap.currentCSSZoom = 0.5     tt.currentCSSZoom = 0.5   (inherited)
tt.getBoundingClientRect() -> w:100 h:30    tt.offsetWidth -> 200
```

So `getBoundingClientRect()` scales **width/height**, not just `left`/`top`. That matters here because `tooltipRect.ttWidth/ttHeight` are sourced from `getBoundingClientRect()` in `Labels.js:52-53` and `Tooltip.js:580`, which makes them viewport px as well.

Two coordinate spaces are in play:

* **V** (viewport): `clientX/Y`, and every `getBoundingClientRect()` field including width/height
* **L** (local CSS px): `style.left/top`, SVG attributes, `w.layout.*`, and literal padding constants

`L = V / zoom`. Every V term needs the division, every L term must stay out of it.

### Blocking: `ttWidth`/`ttHeight` are treated as local px in 3 of the 6 sites

`Position.js` (followCursor path):

```js
x = (ttCtx.e.clientX - seriesBound.left) / zoom
if (x > w.layout.gridWidth / 2) { x = x - ttW }   // subtracts 50 where 100 is needed
```

`x` is now L but `ttW` is still V, so at `zoom: 0.5` the left-flipped tooltip lands half its width off and sits under the cursor. Same shape of issue in `Tooltip.js` `nonAxisChartsTooltips` (where `- tooltipRect.ttWidth / 2` was deliberately pulled outside the division) and in `Intersect.js` `handleMarkerTooltip` (`? ttCtx.tooltipRect.ttWidth : 0`).

The diff is self-contradicting on this point: `moveStickyTooltipOverBars` and the `handleBarTooltip` column path *do* divide `ttHeight` (correct), while the three above do not. Both conventions can't be right.

### Blocking: `w.layout.translateY` is being divided by zoom

In `Position.computeTooltipPosition` and `Intersect.js` (~line 269):

```js
y = (ttCtx.e.clientY + w.layout.translateY - seriesBound.top) / zoom
```

`translateY` is an SVG-local (L) value, so it has to be added after the conversion:

```js
y = (ttCtx.e.clientY - seriesBound.top) / zoom + w.layout.translateY
```

That's 20px of error at `zoom: 0.5` with the default `translateY: 20`, and it grows with top padding / title height. The new test asserts `(120 + 20 - 10) / 0.5 = 260`; the correct value is `240`, so the test currently locks the bug in.

### Minor

**Literal paddings get divided too.** In `handleBarTooltip`, `(clientX - seriesBound.left + 15) / zoom` turns the intended 15px cursor gap into 30px at `zoom: 0.5`, and likewise for the `- 15` in the column branch. Those constants belong outside the division.

**Suggested restructure.** Rather than six copies of `w.dom.elWrap.currentCSSZoom || 1`, I'd normalize once at measurement time: divide `ttWidth`/`ttHeight` by zoom where they're captured (`Labels.js`, `drawFixedTooltipRect`, `updateDimensionCache`), or just read `offsetWidth`/`offsetHeight`, which are already local px per the probe above. Everything downstream then stays in L, each call site needs only a single division on the `clientX - bound.left` term, and both blocking issues above become structurally impossible. If you'd rather keep it inline, a single `ttCtx.getCSSZoom()` helper would still be preferable to six reads.

### Scope

The fix is partial in a way users will notice, since the tooltip ends up correct while its own crosshair does not. Still uncompensated:

* `Position._datapointCenterXFromBars` returns a V-space `bcx`, which feeds `moveXCrosshairs` and the shared-bar tooltip x
* `AxesTooltip.js:209` multiplies a V delta by `yRatio` (an L-derived ratio), so the y-axis tooltip displays a **wrong value**, and line 217 passes a V delta to `moveYCrosshairs`
* `tooltip/Utils.getNearestValues` mixes the L `barPadForNumericAxis` into a V `hoverX`
* Same root cause outside tooltips: `ZoomPanSelection.js`, `ContextMenu.js`, `InkLayer.js`, `Measure.js`, `SVGDraggable.js`

Landing followCursor first is a reasonable slice, but let's frame the PR that way instead of `Closes #5043`. The crosshair mismatch is in the same files, so it would be good to take it in this pass.

### Tests

The 47 tests in `tooltip-positioning.spec.js` pass on the branch. That said, both new tests assert hand-computed arithmetic against a stubbed `getBoundingClientRect`, so they mirror whatever the implementation does and would pass with all of the above intact. The second one (`delete w.dom.elWrap.currentCSSZoom` on a property that was never set) only exercises the `|| 1` fallback.

What would actually catch V/L mixing is a Playwright case in `tests/interaction/`: wrap a chart in `zoom: 0.5`, hover, and assert the tooltip's real `getBoundingClientRect()` sits at the cursor. Could you add one? Ideally covering the marker, bar, and non-axis (pie) paths, since each takes a different branch.

### Looks good

No `Options.js` change, so no `types/apexcharts.d.ts` update is required here. The `|| 1` fallback keeps browsers without `currentCSSZoom` on today's behavior instead of regressing them, which is the right call. Worth a line in the PR description noting that limitation.

Happy to take another look once the coordinate-space handling is consistent.

PR #5271, opened 2026-08-10
fix(annotations): measure the label background from the grid origin (closes #5270)

### LINKED ISSUE #5270: y-axis annotation's label background rectangle is drawn to the right of the label text it belongs behind
### Description

On a `datetime` x-axis, a y-axis annotation's label **background rectangle is drawn to the right of the label text it belongs behind** — by exactly the number of pixels the grid group extends to the left of its own origin. On the attached reproduction that is **20.5px** on an 800×400 chart. Where the label text is lighter than the page (white on a coloured background is the common case), the characters left sticking out of the box paint straight onto the page, and the whole thing reads as a background clipped on its left edge.

`src/modules/annotations/Helpers.js`, `addBackgroundToAnno()` draws the rect in the annotation group's **local** coordinates, but measures the offset to it from the grid group's **rendered** left edge:

```js
const x1 = (coords.left - elGridRect.left) / zoom - pleft
```

Those are the same point only while `.apexcharts-grid` draws nothing outside its own box. On a datetime axis it routinely draws outside: `utils/DateTime.js` `ceilToBoundary()` returns the calendar boundary at or *before* the first data point when the picked interval has step 1 —

```js
case 'year': {
  const y = d.getFullYear()
  const aligned = Math.ceil(y / step) * step
  if (aligned === y && d.getMonth() === 0 && d.getDate() === 1 && ...)
    return timestamp
  return new Date(aligned, 0, 1).getTime()   // floor, not ceiling
}
```

— with `step === 1`, `aligned === y` always, so a first point that is not already midnight on 1 January comes back as 1 January **of its own year**, before `minX`. `TimeScale.generateBaseTicks()` gives that tick `position = (t - minX) / span * gridWidth`, a negative number, and `axes/Grid.js` `datetimeLines()` draws the gridline there verbatim. The grid group's bbox therefore starts at a negative `x`, `elGridRect.left` is no longer the local origin, and every annotation background is displaced by the difference. `month`, `hour`, `minute` and `second` have the same shape.

**It needs five things at once**, which is worth stating because four of themmare easy to lose while minimizing a repro:

1. the picked interval has **step 1** — `pickInterval()` takes the ladder entry nearest `span / 10` in log space, so a 10-to-15-year span picks `year × 1`  and fires, while a 20-year span picks `year × 2` and does not (`ceil(2013/2)*2 = 2014`, after `minX`);
2. the first data point is not on that unit's boundary;
3. **`grid.xaxis.lines.show: true`** — this is not the default. The overhanging x-axis *label* and *tick* are both skipped explicitly (`axes/XAxis.js` drops a label with `label.x + halfWidth < 0`, and `drawXaxisTicks()` returns early on `x1 < 0`), so the vertical gridline is the only one of the three that renders — and the only one inside `.apexcharts-grid`, which is the element being measured. With the default `false` the bbox stays at 0 and nothing is visible;
4. **a small overhang.** `dimensions/XAxis.js` sets `skipFirstTimelinelabel` when `firstTimescaleLabel.position - lbWidth / 1.75 + yAxisWidthLeft < -10`, and `Grid.js` then declines to draw that gridline at all. So a *bigger* overhang stops reproducing — a start date further from the boundary is a weaker repro, not a stronger one;
5. not a bar chart on a numeric axis (`barPadForNumericAxis === 0`) — see the PR, where the existing code compensates for that overhang separately.

### Steps to Reproduce

1. Open the reproduction. It renders two `area` charts side by side, identical in every respect except the first date, and prints its own measurements to the page and the console.
2. **Chart A** — 159 monthly points starting **2013-05-21** (a 13.2-year span, so the tick interval is 1 year), `xaxis.type: 'datetime'`, `grid.xaxis.lines.show: true`, `yaxis.labels.minWidth: 60`, and one y-axis annotation whose label has `position: 'left'` and a background colour.
3. **Chart B** — the control: the same 159 points starting **2013-01-01**, on the boundary, so `ceilToBoundary()` hits its early return.
4. Read `grid bbox x` first, then `box - text` for each chart.

### Expected Behavior

The background rectangle encloses its label text on both charts: `box left - text left` ≈ **−5px**, the label's left padding. Chart B, whose grid does not overhang, gives exactly that.

### Actual Behavior

Chart A puts the background **+15.5px** to the right of its text — a 20.5px displacement — while chart B is correct.

| | chart A (starts 21 May 2013) | chart B (starts 1 Jan 2013) |
|---|---|---|
| `grid.getBBox().x` | **−20.5** | 0.0 |
| first tick wants x | −21.1 | 0.0 |
| leftmost gridline | −20.5, drawn | drawn at 0, as the axis border |
| box left | 94.6 | 74.1 |
| text left | 79.1 | 79.1 |
| **box left − text left** | **+15.5** | −5.0 |

The two measurements are independent and they agree: `15.5 − (−5) = 20.5`, which is `grid.getBBox().x` to the tenth. The background is displaced by exactly the term the expression fails to account for.

**Version:** 6.8.0 (npm `latest`). Also reproduced on 6.7.0 and 6.7.1, with numbers identical to the decimal — `annotations/Helpers.js`, `utils/DateTime.js` and `modules/TimeScale.js` are byte-identical across all three releases.

**Affected versions:** v5 and v6. **v4 and earlier are not affected**: the older `TimeScale.js` (`generateYearScale` / `determineInterval`, no `ceilToBoundary`) computes its first tick as `remainingDaysOfFirstYear * daysWidthOnXAxis`, a *positive* offset to the next boundary, so the grid never overhangs by this route. Worth knowing before testing a pen forked from an old template — it will come back clean whatever the annotation config says.

### Screenshots

Image

### Reproduction Link

[CodePen Reproduction](https://codepen.io/editor/michael-rash/pen/019fe7f3-8cfe-76bb-9188-75168f266089)

mrash:
`addBackgroundToAnno()` draws an annotation's label background in the annotation group's LOCAL coordinates, but measured the offset to it from the grid group's RENDERED left edge. Those are the same point only while `.apexcharts-grid` draws nothing outside its own box, and it draws outside on both axis types this function serves, so the background lands to the right of its own text by whatever the grid overhangs.

On a datetime axis, `DateTime.js` `ceilToBoundary()` floors a step-1 interval to the calendar boundary at or before the first data point, so `TimeScale` gives that tick a negative position and `Grid.js` `datetimeLines()` draws the gridline there verbatim. A 13-year monthly series starting 21 May 2013 puts a year line 20.5px left of the plot on an 800px chart, and every annotation background is drawn 20.5px right of the text it belongs behind. Since label text is commonly light on a coloured background, the characters that fall outside paint onto the page and the whole thing reads as a background clipped on its left edge.

`AxisMapping`'s header and `ZoomPanSelection._screenXToPlotPx` already name this hazard: measuring from the `.apexcharts-grid` box and subtracting `barPadForNumericAxis` is "a fragile pair that only cancels while the grid box happens to extend exactly barPad". The datetime overhang is the predicted failure: the grid box stops extending exactly `barPad`, so the pair stops cancelling.

Two other sites still pair the same two corrections and are **left for a follow-up** (tracked separately, out of scope here):

- `tooltip/Utils.js` `getNearestValues()`, where the grid element rect supplies both the origin *and* the `hoverWidth` divisor, so the same overhang skews the hover hit-test near the plot edges;
- `tooltip/Position.js` `_datapointCenterXFromBars()`.

Derive the origin from the grid's own bbox, which is already in scope for the zoom factor, and drop the `barPad` term rather than moving it: `bbox.x` is the measured overhang where `barPad` is an assumed one, so keeping both would double-count and shift the box a bar-width left on a numeric bar chart that renders correctly today. Where `bbox.x` is 0 the arithmetic is unchanged, so no chart that renders correctly can move.

The new form is also the correct one in two cases the old pair got wrong in the *other* direction, both of which it now fixes:

- a numeric bar chart with `grid.yaxis.lines.show: false` draws no widened horizontal gridline, so `bbox.x` is 0 while `barPad` is not;
- a category axis with `padHorizontal > 0` and yaxis lines off has a positive `bbox.x`.

Both were previously drawn left of their text.

Adds a unit test that overrides the `getBBox` stub in `tests/unit/setup.js`, which returns `x: 0` -- precisely the value at which measuring from either point gives the same answer, and the reason the existing coverage could not have caught this. The e2e spec `line-with-annotations-zoom-level` already asserts the exact invariant (`rect.x <= label.x + maxPadding`); it passes today because its sample leaves `grid.xaxis.lines.show` at the default `false`, so nothing renders at the floored tick and the grid bbox stays at 0.

Fixes #5270

## Type of change

## Checklist:

### DIFF
--- a/src/modules/annotations/Helpers.js
+++ b/src/modules/annotations/Helpers.js
@@ -79,10 +79,22 @@ export default class Helpers {
       ;[ptop, pbottom, pleft, pright] = [pleft, pright, ptop, pbottom]
     }
 
-    const x1 = (coords.left - elGridRect.left) / zoom - pleft
-    const y1 = (coords.top - elGridRect.top) / zoom - ptop
+    // The rect is drawn in the annotation group's LOCAL coordinates, so these
+    // client-space deltas have to be measured from the grid's local (0, 0) --
+    // not from the left/top edge of whatever the grid group happens to render.
+    // Those are the same point only while the grid draws nothing outside its
+    // own box, and it draws outside on both axis types this file serves: a
+    // datetime axis whose first timescale tick is floored to the calendar
+    // boundary before minX (DateTime.js ceilToBoundary), and a numeric bar
+    // chart, whose gridlines run out to -barPadForNumericAxis so edge bars are
+    // not clipped. bbox.x/y are in local units, hence the zoom multiply.
+    const gridLeft = elGridRect.left - gridBBox.x * zoom
+    const gridTop = elGridRect.top - gridBBox.y * zoom
+
+    const x1 = (coords.left - gridLeft) / zoom - pleft
+    const y1 = (coords.top - gridTop) / zoom - ptop
     const elRect = this.annoCtx.graphics.drawRect(
-      x1 - w.globals.barPadForNumericAxis,
+      x1,
       y1,
       coords.width / zoom + pleft + pright,
       coords.height / zoom + ptop + pbottom,
--- a/tests/unit/annotation-helpers.spec.js
+++ b/tests/unit/annotation-helpers.spec.js
@@ -160,6 +160,68 @@ describe('Annotation Helpers', () => {
       const result = helpers.addBackgroundToAnno(annoEl, anno)
       expect(result).toBeDefined()
     })
+
+    it('measures from the grid origin, not the grid element edge', () => {
+      // The rect is drawn in the annotation group's local coordinates, so the
+      // offset to it has to be measured from the grid's local (0, 0). That is
+      // NOT the grid element's rendered left edge whenever the grid draws
+      // outside its own box -- a datetime axis whose first timescale tick is
+      // floored to before minX, or a numeric bar chart whose gridlines run out
+      // to -barPadForNumericAxis.
+      //
+      // getBBox is stubbed to x: 0 in tests/unit/setup.js, which is precisely
+      // the value at which measuring from either point gives the same answer,
+      // so it has to be overridden here or this test cannot fail.
+      chart = createChartWithOptions({
+        chart: { type: 'line' },
+        series: [{ data: [10, 20, 30, 40, 50] }],
+        annotations: {
+          yaxis: [{ y: 30, label: { text: 'Label', position: 'left' } }],
+        },
+      })
+      const annoCtx2 = new Annotations(chart.w)
+      const helpers2 = new Helpers(annoCtx2)
+
+      const baseEl = chart.w.dom.baseEl
+      const gridEl = baseEl.querySelector('.apexcharts-grid')
+      const annoEl = baseEl.querySelector('.apexcharts-yaxis-annotation-label')
+
+      // The grid renders 20px left and 5px above its own origin, so that origin
+      // sits at client (100, 15) while the element's edge is at (80, 10).
+      vi.spyOn(gridEl, 'getBBox').mockReturnValue({
+        x: -20,
+        y: -5,
+        width: 720,
+        height: 400,
+      })
+      vi.spyOn(gridEl, 'getBoundingClientRect').mockReturnValue({
+        left: 80,
+        top: 10,
+        width: 720, // width / bbox width => a zoom of 1
+        height: 400,
+      })
+      vi.spyOn(annoEl, 'getBoundingClientRect').mockReturnValue({
+        left: 150,
+        top: 30,
+        width: 40,
+        height: 12,
+      })
+
+      const anno = {
+        label: {
+          text: 'Label',
+          style: { padding: { left: 4, right: 4, top: 2, bottom: 2 } },
+        },
+      }
+
+      const elRect = helpers2.addBackgroundToAnno(annoEl, anno)
+
+      // (150 - 100) - 4 = 46, and (30 - 15) - 2 = 13. Measuring from the
+      // element edge instead gives 66 and 18: the background drawn a full
+      // overhang right of, and below, the text it belongs behind.
+      expect(Number(elRect.node.getAttribute('x'))).toBeCloseTo(46, 5)
+      expect(Number(elRect.node.getAttribute('y'))).toBeCloseTo(13, 5)
+    })
   })
 
   describe('getY1Y2', () => {


junedchhipa (approved):
Verified the premise end to end and the fix is correct.

- `.apexcharts-grid` and the annotation groups are both direct children of `elGraphical` with no transform of their own, so the rect's local space and the grid's `getBBox()` space really are the same space. `elGridRect.left - bbox.x * zoom` is therefore the client coordinate of local `x = 0` for **any** value of `bbox.x`; the old formula was right only when `bbox.x === -barPad`.
- Ordering is safe: `drawAxesAnnotations()` runs after the grid group is added in both `grid.position` branches, so the bbox is populated when `addBackgroundToAnno()` reads it.
- The new test genuinely fails without the src change (`expected 66 to be close to 46`) -- confirmed by reverting `Helpers.js` to main and re-running.
- 105 annotation unit tests pass. `npm run lint` is clean. `npm run typecheck` adds no new errors (24 pre-existing, all in `Drilldown.js`). The prettier drift on both touched files is pre-existing on `main`.
- No e2e snapshot churn: `line-with-annotations-zoom-level` is a line chart (`barPad === 0`) that never sets `grid.xaxis.lines.show`, so the grid group holds only the yaxis horizontal lines from `x1 = 0` plus `drawGridArea`'s two transparent lines at `padHorizontal === 0`. `bbox.x` is 0 and the arithmetic is byte-identical.

Two notes, both handled rather than blocking:

1. "This was the last place doing that" was not accurate -- `tooltip/Utils.js` `getNearestValues()` and `tooltip/Position.js` `_datapointCenterXFromBars()` still pair the grid rect with `-barPadForNumericAxis`, and in the first the grid rect supplies the `hoverWidth` divisor as well. Reworded in the commit message and PR body, and tracked in #5272.
2. The gridline that causes this is still drawn, unclipped, into the y-axis gutter (`_drawGridLine()` has no clip-path -- only grid bands get `gridRectMask`). Fixing the annotation geometry so it does not depend on what the grid draws is the right call regardless, but the repro chart still shows a stray line. Tracked in #5273.

Also amended: the test now assigns to the shared `chart` so `afterEach` disposes it, instead of a local `chart2` destroyed at the end of the body, which leaked the chart and its ResizeObserver into later tests whenever an assertion failed.

Worth recording that the new form also *fixes* two cases the old pair got wrong in the other direction: a numeric bar chart with `grid.yaxis.lines.show: false` (no widened horizontal gridline, so `bbox.x === 0` while `barPad` is not) and a category axis with `padHorizontal > 0` and yaxis lines off (positive `bbox.x`). Both were previously drawn left of their text. Added to the commit message.

Nice diagnosis, and thank you for the write-up in #5270 -- the five preconditions were exactly right.

PR #5274, opened 2026-08-10
fix(grid): skip datetime gridlines that fall outside the plot (fixes #5273)

### LINKED ISSUE #5273: Datetime axis draws an unclipped gridline left of the plot, in the y-axis gutter
Split out of #5270. That issue reported the visible *consequence* (annotation label backgrounds displaced right), fixed by #5271 by making the annotation geometry independent of what the grid draws. The gridline that caused it is still drawn, and it is visible on its own.

### What happens

`utils/DateTime.js` `ceilToBoundary()` returns the calendar boundary at or *before* the first data point when the picked interval has step 1:

```js
case 'year': {
  const y = d.getFullYear()
  const aligned = Math.ceil(y / step) * step
  if (aligned === y && d.getMonth() === 0 && d.getDate() === 1 && ...)
    return timestamp
  return new Date(aligned, 0, 1).getTime()   // floor, not ceiling
}
```

With `step === 1`, `aligned === y` always, so a first point that is not already on the unit boundary comes back *before* `minX`. `TimeScale.generateBaseTicks()` gives that tick a negative `position`, and `axes/Grid.js` `datetimeLines()` draws the gridline there verbatim. `month`, `hour`, `minute` and `second` have the same shape.

The overhanging x-axis **label** and **tick** are both suppressed (`AxesUtils` drops a label with `label.x + halfWidth < 0`; `drawXaxisTicks()` returns early on `x1 < 0`), but the gridline is not, and `_drawGridLine()` attaches no clip-path -- only grid *bands* get `gridRectMask` (`axes/Grid.js`, `_drawGridBandRect`). So the line renders unclipped into the y-axis label gutter, left of the plot area.

### Repro

Same chart as #5270: 159 monthly points starting 2013-05-21 (13.2-year span, so the interval is year x 1), `xaxis.type: 'datetime'`, **`grid.xaxis.lines.show: true`**, `yaxis.labels.minWidth: 60`, 800x400. The stray line sits 20.5px left of the plot origin, over the y-axis labels.

Note the constraints from #5270 still apply: it needs a step-1 interval, a first point off the boundary, `grid.xaxis.lines.show: true`, and a *small* overhang (a large one trips `skipFirstTimelinelabel` in `dimensions/XAxis.js`, after which `Grid.js` declines to draw the line at all).

### Two candidate fixes

1. **`ceilToBoundary()`** -- make the step-1 path actually ceil, so no tick is generated before `minX`. Most direct, but it changes the first timescale tick on every datetime chart, so it needs an e2e sweep.
2. **`datetimeLines()` / `_drawGridLines()`** -- skip a tick whose `position < 0` (and `> gridWidth`), matching what the label and tick paths already do independently. Narrower, and arguably where the invariant belongs, since the drawing code is what has the plot bounds.

Option 2 is the safer first move; option 1 is worth investigating separately since a tick before `minX` may be skewing other consumers.

lazerg:
On a step-1 datetime interval, `DateTime.ceilToBoundary()` returns the calendar boundary at or *before* the first data point, so `TimeScale.generateBaseTicks()` emits a tick with a negative position. The x-axis label for that tick is dropped (`label.x + halfWidth < 0`) and `drawXaxisTicks()` returns early on it, but `Grid.datetimeLines()` passed the position straight to `_drawGridLine()`. Gridlines carry no clip-path, only grid bands get `gridRectMask`, so the line rendered in the y-axis label gutter, left of the plot.

The loop now applies the same bounds check `drawXaxisTicks()` already uses, so a tick outside the plot no longer contributes a line. Ticks inside the plot are untouched, and a chart whose first tick already lands at or after `minX` renders exactly as before, which the e2e sweep confirms.

Fixes #5273

## Type of change

## Checklist:

### DIFF
--- a/src/modules/axes/Grid.js
+++ b/src/modules/axes/Grid.js
@@ -263,6 +263,8 @@ class Grid {
         x1 = /** @type {any} */ (this.xaxisLabels[i]).position
         x2 = /** @type {any} */ (this.xaxisLabels[i]).position
 
+        if (x1 < 0 || x1 - 2 > w.layout.gridWidth) continue
+
         this._drawGridLines({
           i,
           x1,
--- a/tests/unit/issue-5273-datetime-gridline-overhang.spec.js
+++ b/tests/unit/issue-5273-datetime-gridline-overhang.spec.js
@@ -1,0 +1,55 @@
+import { describe, it, expect } from 'vitest'
+import { createChartWithOptions } from './utils/utils.js'
+
+/**
+ * #5273: on a step-1 datetime interval, `DateTime.ceilToBoundary()` returns the
+ * calendar boundary at or *before* the first data point, so `TimeScale` gives
+ * that tick a negative position. The label and the axis tick are both dropped
+ * for being outside the plot, but `Grid.datetimeLines()` drew the gridline
+ * verbatim, and gridlines get no clip-path, so the line landed in the y-axis
+ * label gutter.
+ *
+ * 159 monthly points from 21 May 2013 span 13.2 years, which picks year x 1.
+ */
+
+const DATA = Array.from({ length: 159 }, (_, i) => [
+  new Date(2013, 4 + i, 21).getTime(),
+  100 + (i % 17),
+])
+
+function render() {
+  return createChartWithOptions({
+    chart: { type: 'line', width: 800, height: 400 },
+    series: [{ name: 'Series', data: DATA }],
+    xaxis: { type: 'datetime' },
+    yaxis: { labels: { minWidth: 60 } },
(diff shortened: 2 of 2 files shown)


junedchhipa (approved):
Reviewed: the guard matches the predicate `XAxis.drawXaxisTicks()` already applies, so the gridline and axis tick now agree. `continue` is safe here since `x1`/`x2` are reassigned each iteration, and no downstream consumer indexes the vertical-gridline group positionally. Verified the new spec fails on `main` and passes with the patch; full unit suite green (2184 tests). Thanks!

PR #5275, opened 2026-08-10
fix(tooltip): measure the hover hit-test and bar centers from the plot origin (closes #5272)

### LINKED ISSUE #5272: Tooltip hover hit-test and bar crosshair center still pair the .apexcharts-grid rect with -barPadForNumericAxis
Follow-up to #5271, which removed the last such pair from `annotations/Helpers.js`. Two sites in the tooltip path still do it, so they carry the same latent bug that #5270 reported for annotation backgrounds.

`AxisMapping`'s header and `ZoomPanSelection._screenXToPlotPx` both already document why this shape is unsafe: measuring from the `.apexcharts-grid` element's client rect and subtracting `barPadForNumericAxis` is "a fragile pair that only cancels while the grid box happens to extend exactly barPad". The grid box stops extending exactly `barPad` whenever the grid draws outside its own box for some *other* reason, and on a `datetime` axis it routinely does (see #5270: `ceilToBoundary()` floors a step-1 interval to the boundary before `minX`, `TimeScale` gives that tick a negative position, and `datetimeLines()` draws the gridline there).

### 1. `tooltip/Utils.js` `getNearestValues()`

```js
const seriesBound = elGrid.getBoundingClientRect()
const hoverWidth = seriesBound.width
...
const hoverX = clientX - seriesBound.left - w.globals.barPadForNumericAxis
```

The grid rect supplies **both** the origin and the `hoverWidth` divisor, so on a chart with a 20.5px overhang the origin is 20.5px too far left *and* `xDivisor` is computed over a width 20.5px too wide. The resolved `j` is wrong near the plot edges, and the `notInRect` guard is measured against the inflated box too.

### 2. `tooltip/Position.js` `_datapointCenterXFromBars()`

```js
return (
  (unionLeft + unionRight) / 2 -
  gridRect.left -
  (w.globals.barPadForNumericAxis || 0)
)
```

Same pair, so the crosshair lands off the bar center by the overhang.

### Suggested fix

Not the bbox derivation used in #5271. There is already a render-independent plot origin in the codebase: `elGraphical` is `translate(translateX, translateY)` (`Core.js` `shiftGraphPosition()`), so local `(0, 0)` in client space is `svgRect.left + translateX` (times the CSS zoom factor if one is in play). That is exactly what `ZoomPanSelection._screenXToPlotPx()` uses, and its doc comment recommends it over the grid box for this reason. Routing both sites through it drops the dependency on what the grid happens to render, and lets the `barPad` term go rather than being re-derived.

Worth auditing whether `hoverWidth` should be `w.layout.gridWidth` at the same time, since that is the value the divisor actually wants.

lazerg:
`getNearestValues()` measured the cursor against the `.apexcharts-grid` element's client rect and subtracted `barPadForNumericAxis`, and `_datapointCenterXFromBars()` did the same to the bar union's center. That pair only cancels while the grid box happens to extend exactly `barPad`, and it stops doing so as soon as the grid draws outside its own box for another reason. A datetime axis does that routinely: `ceilToBoundary()` floors a step-1 interval to the boundary before `minX`, `TimeScale` gives that tick a negative position, and `datetimeLines()` draws the gridline there. On the 13-year monthly series from #5270 that is a 20.5px overhang on an 800px chart.

In `getNearestValues()` the grid rect supplied both the origin and the `hoverWidth` divisor, so the hover x came out an overhang too large while `xDivisor` was computed over a box that much too wide. The resolved `j` was wrong near the plot edges and the `notInRect` guard let in points that sat left of the plot. In `_datapointCenterXFromBars()` the crosshair landed off the bar center by the same amount.

Both now go through `AxisMapping.screenXToPlotPx()`, which anchors on the svg element's left edge plus `translateX`, the origin the bars are laid out from, and scales by the container's CSS zoom. `ZoomPanSelection._screenXToPlotPx()` delegates to it rather than keeping a second copy of the same derivation, and its doc comment already recommended that reference over the grid box. `hoverWidth` becomes `w.layout.gridWidth`, which is the value the divisor actually wants, so the vestigial `hoverX * (gridRect.width / hoverWidth)` rescale ahead of `closestInMultiArray` (always a no-op, since both terms came off the same element) goes away. The bounds check in `handleStickyTooltip` moves to `gridWidth` for the same reason.

Unit tests cover both sites with a grid element whose rect starts 20.5px before the plot origin. All 2185 unit tests and the 211 Playwright interaction tests pass.

Fixes #5272

## Type of change

## Checklist:

### DIFF
--- a/src/modules/AxisMapping.js
+++ b/src/modules/AxisMapping.js
@@ -27,7 +27,8 @@
  * space the series `<g transform="translate(translateX, translateY)">` and the
  * selection rect's transform live in), so a value produced by
  * {@link AxisMapping.dataXToPx} can be used directly as an SVG `x` attribute and
- * a value fed to {@link AxisMapping.pxToDataX} is `screenX - svgLeft - translateX`.
+ * a value fed to {@link AxisMapping.pxToDataX} is what
+ * {@link AxisMapping.screenXToPlotPx} returns.
  *
  * @module modules/AxisMapping
  */
@@ -61,5 +62,24 @@ export default class AxisMapping {
    */
   static pxToDataX(w, px) {
     return w.globals.minX + px * AxisMapping.xRatio(w)
+  }
+
+  /**
+   * Client (screen) x -> pixels from the plot origin. The origin is the svg
+   * element's left edge plus `translateX`, never the `.apexcharts-grid` box
+   * (fact 2 above), so the result does not depend on what the grid happens to
+   * render. `svgWidth` is the unscaled width the svg was drawn at, so the ratio
+   * against the measured one is the CSS zoom of any container the chart sits in.
+   * @param {import('../types/internal').ChartStateW} w
+   * @param {number} screenX
+   * @returns {number}
+   */
+  static screenXToPlotPx(w, screenX) {
+    const baseEl = w.dom.baseEl
+    const svg = baseEl && baseEl.querySelector('.apexcharts-svg')
+    if (!svg) return screenX - w.layout.translateX
+    const svgRect = svg.getBoundingClientRect()
+    const zoom = w.globals.svgWidth ? svgRect.width / w.globals.svgWidth : 1
+    return (screenX - svgRect.left) / (zoom || 1) - w.layout.translateX
   }
 }
--- a/src/modules/ZoomPanSelection.js
+++ b/src/modules/ZoomPanSelection.js
@@ -1268,10 +1268,7 @@ export default class ZoomPanSelection extends Toolbar {
    * @returns {number}
    */
   _screenXToPlotPx(screenX) {
-    const baseEl = this.w.dom.baseEl
-    const svg = baseEl && baseEl.querySelector('.apexcharts-svg')
-    const svgLeft = svg ? svg.getBoundingClientRect().left : 0
-    return screenX - svgLeft - this.w.layout.translateX
+    return AxisMapping.screenXToPlotPx(this.w, screenX)
   }
 
   /**
--- a/src/modules/tooltip/Position.js
+++ b/src/modules/tooltip/Position.js
@@ -1,4 +1,5 @@
 // @ts-check
+import AxisMapping from '../AxisMapping'
 import Graphics from '../Graphics'
 import Series from '../Series'
 import { ARROW_TIP_OVERHANG } from './constants'
@@ -661,7 +662,7 @@ export default class Position {
       // for odd-count series. Use the bar's rendered DOM rect instead so
       // the data-point center is correct regardless of stack/group layout.
       if (jBar && !isBoxOrCandle) {
-        const center = this._datapointCenterXFromBars(j, seriesBound)
+        const center = this._datapointCenterXFromBars(j)
         if (center != null) {
           bcx = center
         } else {
@@ -730,10 +731,9 @@ export default class Position {
    * attribute math in `moveStickyTooltipOverBars`. Returns null when no
    * usable bars are found.
    * @param {number} j
-   * @param {DOMRect} gridRect
    * @returns {number | null}
    */
-  _datapointCenterXFromBars(j, gridRect) {
+  _datapointCenterXFromBars(j) {
     const w = this.w
     const bars = w.dom.baseEl.querySelectorAll(
       `.apexcharts-bar-series path[j='${j}'],` +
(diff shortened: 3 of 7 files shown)


junedchhipa:
Nice fix, and thanks for consolidating the derivation into `AxisMapping` instead of adding a third copy. I reviewed this in the browser against 6.8.0 and pushed one commit on top (`60bf1d6`) rather than leaving you a round trip. Details below so nothing is a surprise, and please push back if you disagree with any of it.

### What I changed

**1. The in/out bound got tightened as a side effect.** Switching `hoverWidth` to `w.layout.gridWidth` is right, but the bound in `handleStickyTooltip` also moved from the grid element's width to `gridWidth`. On a numeric or datetime axis with bars those are not the same number: `Grid._drawGridLine` widens the horizontal gridlines group by `barPadForNumericAxis` on each side, so the element measures `gridWidth + 2 * barPad`.

That matters because bars at the first and last data point are centred on grid-local `0` and `gridWidth`, so each one reaches half a bar width outside the plot box. On a 5-point datetime column chart (`shared: true, intersect: false`): `gridWidth` 498.8, last bar spans grid-local 457.0 to 540.7. Between 498.8 and 540.7 you are still on the bar, but the new bound sends `handleStickyTooltip` down its `handleMouseOut` path. I widened it by the pad on both sides, and made the same allowance behind the zoom/pan cursor class in `getNearestValues`, which shared the bound.

Calibrating the severity honestly: I could **not** get this to produce a visible symptom. In every config I built, the tooltip still ends up active over that overhang because other listeners redraw it after `handleMouseOut` runs. So treat this as restoring tolerance the old code had rather than as a user-facing bug fix. It is cheap and it removes an unmotivated behaviour change, which is why I pushed it, but I would not have blocked the PR on it.

**2. Test fixture.** `makeCtx` in `tooltip-positioning.spec.js` builds `baseEl` as a bare `div`, so the three re-based `_datapointCenterXFromBars` tests had no `.apexcharts-svg` to find and silently ran `screenXToPlotPx`'s `!svg` fallback. Their expected values are unchanged only because `svgLeft` is 0 there, so the comments ("plot origin = svgLeft (0) + translateX (50)") describe a path they do not execute, and they would stay green if the svg/zoom derivation broke. I gave `makeCtx` a measurable svg and dropped the now-duplicate local stub from your new test.

I added two unit tests for the bound; both fail without the change and pass with it. `2188` unit tests and `211` interaction tests pass.

### Two things for you

**The stated root cause does not hold, and the description undersells the real one.** There is no `x1 < 0` guard in `_drawGridLines`, so a floored boundary tick can draw at a negative position, but whether it does depends on `skipFirstTimelinelabel`, which `dimensions/XAxis.js` sets from label-text overflow rather than tick position. On the 13-year monthly series from #5270 I measured `.apexcharts-grid`'s left at exactly `origin - barPad`, so the two old corrections cancelled and base and PR returned identical `hoverX`/`j`.

The half of the bug that is always present is the `xDivisor` inflation: `hoverWidth` was `gridWidth + 2 * barPad`, so `j = Math.ceil(hoverX / xDivisor) - 1` drifted on every numeric bar chart regardless of what the grid drew. That is the fix worth naming. It is also demonstrable when the grid does not draw the pad widening: with `grid: { yaxis: { lines: { show: false } } }` on a datetime column chart the old code returned `hoverX = -3.12` at the plot origin and resolved `j = 154` instead of `155`; this branch returns `0` and `155`. Could you reword the description along those lines so the changelog reflects the mechanism that actually fires?

**`hoverY` is now on a different scale from `hoverX`.** `hoverX` goes through the CSS-zoom normalization but `hoverY` is still `clientY - seriesBound.top`, so `closestInMultiArray` mixes scales. Under `transform: scale(0.5)` I measured hoverX 488.1 (correct) against hoverY 112.2 where the true value is 225.5, which under-weights Y in series capture and scatter/bubble `j`. This is pre-existing rather than caused by your change, and fixing it properly means auditing the other screen-to-plot Y conversions, so I would keep it out of this PR. I will open a separate issue unless you would rather take it.

Also worth a line in the description: routing `ZoomPanSelection._screenXToPlotPx` through the shared helper makes selection/zoom geometry zoom-aware, and the `dragX` 10px threshold is now measured in unscaled chart px. Both look like improvements, but they are behaviour changes inside a CSS-scaled container that a reader would not expect from the title.

junedchhipa (approved):

PR #5276, opened 2026-08-10
fix(annotations): clamp point annotation label x within chart bounds (Fixes #5106)

### LINKED ISSUE #5106: Point annotation labels are cut of
### Description

Point annotations near left or right chart boundary are cut of with a longer text.

### Steps to Reproduce

1. Add point annotations with longer text at left / right boundary

### Expected Behavior

Labels of annotations should respect the chart boundaries and should be moved into the chart. 

### Reproduction Link

waterWang:
## Description

Fixes #5106 — Point annotation labels near the left or right chart boundary are cut off when the label text extends beyond the chart area.

### Problem

When a point annotation has a long label text and is positioned near the left or right edge of the chart, the label extends beyond the chart's grid bounds and gets clipped (hidden) by the SVG viewport.

### Fix

Measure the label text width using  before rendering, then clamp the label x position to keep it within the chart's grid boundaries:

- **textAnchor = "start"** (default): clamp x to [0, gridWidth - labelWidth]
- **textAnchor = "end"**: clamp x to [labelWidth, gridWidth]
- **textAnchor = "middle"**: clamp x to [labelWidth/2, gridWidth - labelWidth/2]

### Test

Added 3 new tests in the boundary-clamping describe block:
- Right-edge label: verify x is clamped to less than gridWidth
- Left-edge label: verify x is clamped to >= 0
- Already-fitting label: verify x is > 0

### DIFF
--- a/src/modules/annotations/PointsAnnotations.js
+++ b/src/modules/annotations/PointsAnnotations.js
@@ -66,8 +66,34 @@ export default class PointAnnotations {
 
       const text = anno.label.text ? anno.label.text : ''
 
+      // Clamp label x to keep it within chart bounds (prevents cutting off
+      // labels near left/right chart boundary — fixes apexcharts/apexcharts.js#5106)
+      const gridWidth = w.layout.gridWidth
+      const labelWidth = text
+        ? this.annoCtx.graphics.getTextRects(
+            text,
+            anno.label.style.fontSize,
+            anno.label.style.fontFamily
+          ).width
+        : 0
+
+      let labelX = x + anno.label.offsetX
+      if (labelWidth > 0) {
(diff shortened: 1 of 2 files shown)


junedchhipa:
Thanks for this, and apologies it sat for a few weeks. You found this before anyone else did and the diagnosis is right.

We are going with #5292, which arrived later but takes a slightly different approach: it shifts the label by exactly the amount it overflows, so labels that already fit are left untouched, and it measures the text at the weight it actually renders at, which matters for bold labels. It also accounts for the label's background box rather than just the text.

Closing this in favour of that one, with credit to you for getting there first. #5106 will close when it merges.

PR #5277, opened 2026-08-15
fix: keep millisecond resolution when x is a Date on a datetime axis

### LINKED ISSUE #5054: Line chart with `datetime` axis loses millisond resolution when x values are dates
### Description

In a line chart, `datetime` axis has millisecond-resolution when `x` is a `Number` but second-resolution when `x` is a `Date`. This results in weird charts as two different Y values are shown at the same x value, resulting in a vertical bar. 

I think `new Date(x).getTime == x` so no precision should be lost by passing dates in a series instead of raw numbers.

I ran into this problem after MS Copilot suggested to use `x: new Date(...)`. After switching to raw numbers everything worked as expected.

### Steps to Reproduce

```html
<!DOCTYPE html>
<html>
<head>
	<meta charset="UTF-8">
	<title>Reproducer</title>
	<script src="https://cdn.jsdelivr.net/npm/apexcharts"></script>
</head>
<body>
	<div id="chart"></div>
	<script>
		const chartOptions = {
			chart: {
				type: 'line',
				height: 500
			},
			series: [
				{
					name: "Number",
					data: [
						{x: 1748924002000, y: 1 },
						{x: 1748924002500, y: 2 },
						{x: 1748924003000, y: 1 }
					]
				},
				{
					name: "Date",
					data: [
						{x: new Date(1748924002000), y: 1 },
						{x: new Date(1748924002500), y: 2 },
						{x: new Date(1748924003000), y: 1 }
					]
				}
			],
			xaxis: {
				type: 'datetime'
			}
		};
		const chart = new ApexCharts(document.querySelector("#chart"), chartOptions);
		chart.render();
	</script>
</body>
```

### Expected Behavior

`{x: 1748924002500, y: 2 }` and `{x: new Date(1748924002500), y: 2 },` should be treated equally.

### Actual Behavior

`{x: new Date(1748924002500), y: 2 },` is rounded (?) to next second.

### Screenshots

Image

### Reproduction Link

https://codepen.io/apexcharts/pen/bxzgZ gave me a 404 and I skipped this step. Full reproducer is above.

aron-intframe:
Closes #5054.

It's the `toString()` in `Data.js`:

```js
this.twoDSeriesX.push(dt.parseDate(x.toString()))
```

`Date.prototype.toString()` only has second resolution, so parsing its output back loses the milliseconds:

```js
new Date(1748924002500).toString()
// 'Tue Jun 03 2025 11:13:22 GMT+0700 (Indochina Time)'   <- .500 gone
```

Which is why two points 500ms apart end up on the same x and draw as a vertical line. A numeric x never reaches that branch, and that's what the reporter noticed when switching to raw numbers fixed it.

A Date already holds the exact epoch, so this reads it directly instead of round-tripping through a string. Note `parseDate(x)` on the line above (the `isXString` branch) already preserves it, since `Date.parse` coerces internally and `getTimeStamp` then does `new Date(x).getTime()` — only the `.toString()` call is lossy.

Added `tests/unit/datetime-date-object-x.spec.js` covering the milliseconds, Date-vs-number parity, and that date strings still parse. Against the unpatched `Data.js` the first two fail and the string one passes, which is the right shape.

### DIFF
--- a/src/modules/Data.js
+++ b/src/modules/Data.js
@@ -331,7 +331,12 @@ export default class Data {
           }
         } else {
           if (cnf.xaxis.type === 'datetime') {
-            this.twoDSeriesX.push(dt.parseDate(x.toString()))
+            // A Date already carries an exact epoch value, so read it
+            // directly. Date.prototype.toString() has second resolution, and
+            // parsing its output back drops the milliseconds silently.
+            this.twoDSeriesX.push(
+              x instanceof Date ? x.getTime() : dt.parseDate(x.toString())
+            )
           } else {
             this.w.axisFlags.dataFormatXNumeric = true
             this.w.axisFlags.isXNumeric = true
--- a/src/modules/Data.js
+++ b/src/modules/Data.js
@@ -309,7 +309,7 @@ export default class Data {
       const isXDate = !isXArr && !!dt.isValidDate(x)
 
       if (isXString || isXDate) {
-        // user supplied '01/01/2017' or a date string (a JS date object is not supported)
+        // user supplied a date string ('01/01/2017') or a JS Date object
         if (isXString || cnf.xaxis.convertedCatToNumeric) {
           const isRangeColumn =
             gl.isBarHorizontal && this.w.axisFlags.isRangeData
@@ -335,7 +335,7 @@ export default class Data {
             // directly. Date.prototype.toString() has second resolution, and
             // parsing its output back drops the milliseconds silently.
             this.twoDSeriesX.push(
-              x instanceof Date ? x.getTime() : dt.parseDate(x.toString())
+              x instanceof Date ? x.getTime() : dt.parseDate(x.toString()),
             )
           } else {
             this.w.axisFlags.dataFormatXNumeric = true
--- a/tests/types/consumer-patterns.ts
+++ b/tests/types/consumer-patterns.ts
@@ -285,6 +285,18 @@ const _violinRaw: ApexCharts.ApexOptions = {
 }
 void _boxRaw; void _violinRaw
 
+// datetime axis: a Date, a timestamp and a date string are all valid x values
+const _dateX: ApexCharts.ApexOptions = {
+  chart: { type: 'line' },
+  xaxis: { type: 'datetime' },
+  series: [
+    { name: 'Date', data: [{ x: new Date(1748924002500), y: 1 }] },
+    { name: 'Timestamp', data: [{ x: 1748924002500, y: 2 }] },
+    { name: 'String', data: [{ x: '2025-06-03T04:13:22.500Z', y: 3 }] },
+  ],
+}
+void _dateX
+
 // SSR statics from the apexcharts/ssr entry
 async function _ssrUse() {
   const html = await ApexChartsClass.renderToHTML({ chart: { type: 'bar' } }, { width: 500 })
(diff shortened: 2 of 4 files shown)


junedchhipa (approved):
Thanks, this is a well diagnosed fix and the write-up made it easy to verify.

Confirmed the root cause: `Date.parse` coerces its argument through `Date.prototype.toString()`, so the round trip really does truncate.

```
Date.parse(new Date(1748924002500))  ->  1748924002000
```

A few things I checked before merging:

* **`getTime()` matches the existing helper.** Since the new branch skips `getTimeStamp`, I wanted to be sure it does not sidestep `xaxis.labels.datetimeUTC`. That option's branch (`new Date(new Date(x).toISOString().substr(0, 25)).getTime()`) turns out to be epoch-identity for in-range dates, verified under a non-UTC `TZ` for both UTC and offset-bearing inputs. So a `Date` and its equivalent string still land on the same value, which is the property the issue was really about.
* **This was the only lossy path.** `handleFormat2DArray` already does `new Date(x).getTime()`, and the sibling `convertedCatToNumeric` branch already passes the `Date` straight into `dt.parseDate`. Line 334 was the one place an x value was routed through a string.
* **The test is a real guard.** Reverting only `Data.js` fails exactly the milliseconds and parity cases (`expected [1748924002000, 1748924002000, …] to deeply equal [1748924002000, 1748924002500, …]`) and leaves the date-string case green, matching your description. Full unit suite passes with the patch.

I pushed one follow-up commit rather than send it back to you, since it is small:

* `ApexAxisChartSeries` declared `x: string | number`, so the exact code this fix enables still failed to compile for TypeScript users. Widened to `string | number | Date` and covered the three x forms in `tests/types/consumer-patterns.ts`.
* The comment above the branch claimed a JS date object is not supported, which your change makes untrue.
* Added the trailing comma prettier wants on the new call. (The two other prettier deviations in `Data.js` are pre-existing and left alone.)

One adjacent thing I deliberately did not fold in, for anyone who finds this later: on a non-datetime axis a `Date` x still goes through `parseFloat(x)` and becomes `NaN`, and `dt.isValidDate(new Date('garbage'))` throws because `parseDate`'s fallback assumes a string. Both predate this PR and belong in their own change.

aron-intframe:
Thanks for the thorough merge review, and for pushing the type widening yourself rather than sending it back — `x: string | number` would indeed have made the fix unusable from TypeScript.

I took the two adjacent items you flagged: **#5279**.

Both still reproduced on `main` (b0dc6af):

```
× reads the Date as its epoch on a numeric axis
    AssertionError: expected [ NaN, NaN ] to deeply equal [ 1748924002000, 1748924003000 ]
× makes isValidDate answer false rather than throw
    AssertionError: expected [Function] to not throw an error but 'TypeError: dateStr.replace is not a f…' was thrown
```

`parseFloat` reads a `Date` through `toString()`, which starts with a weekday, so the branch that had just set `dataFormatXNumeric` and `isXNumeric` got `NaN`. And `parseDate`'s salvage pass calls `.replace()` on whatever it was handed, so an invalid `Date` threw out of `isValidDate` and took the render down instead of falling through to the category path.

I reverted each of the two fixes on its own to confirm neither test is vacuous (2 failed / 3 passed each way, 5 passed with both), kept the pre-existing prettier deviations in those files untouched as you did, and checked that `typecheck` reports the same 23 pre-existing `Drilldown.js` errors with and without the change. Full unit suite is 3181 passing.

PR #5279, opened 2026-08-27
fix(data): a Date x on a non-datetime axis, and an invalid Date

aron-intframe:
Picking up the two adjacent problems you named at the end of your review of #5277 and deliberately did not fold in:

> One adjacent thing I deliberately did not fold in, for anyone who finds this later: on a non-datetime axis a `Date` x still goes through `parseFloat(x)` and becomes `NaN`, and `dt.isValidDate(new Date('garbage'))` throws because `parseDate`'s fallback assumes a string. Both predate this PR and belong in their own change.

Both still reproduce on `main` (b0dc6af).

### 1. A `Date` x on a non-datetime axis becomes `NaN`

`Data.js` sets `dataFormatXNumeric` and `isXNumeric` and then reads the value with `parseFloat(x)`. `parseFloat` stringifies its argument first, and `Date.prototype.toString()` starts with a weekday name:

```js
new Date(1748924002000).toString()   // 'Tue Jun 03 2025 11:13:22 GMT+0700 (Indochina Time)'
parseFloat(new Date(1748924002000))  // NaN
```

Those two lines have already decided the x is numeric, and a `Date`'s numeric value is its epoch — the same value the `datetime` branch three lines above reads. So this is the same shape as the fix in #5277, applied to the branch that was left out.

### 2. An invalid `Date` throws instead of answering "not a date"

`isValidDate` hands its argument straight to `parseDate`. For an invalid `Date`, `Date.parse` returns `NaN`, so it falls into the salvage pass, which calls `.replace()` on it:

```js
Date.parse(new Date('garbage'))   // NaN  → falls through to the salvage pass
// → TypeError: dateStr.replace is not a function
```

`Data.js:309` calls `isValidDate(x)` for every x, so a single bad point takes the whole render down with a `TypeError`. Only a string has an alternate spelling worth trying, so anything else now returns `NaN` and `isValidDate` can answer `false`.

**Correction (per @junedchhipa's review):** an earlier draft of this paragraph said the bad x then falls through to the category path. It does not. Failing `isValidDate` means it misses the `isXString || isXDate` branch entirely and lands in the final `else`, which pushes `x` unconverted and sets `isXNumeric` / `dataFormatXNumeric`, so `fallbackToCategory` is never set and `categoryLabels` stays empty. Measured on a category axis: `raw pushed = [object Date]`, `categoryLabels = []`, 0 x axis label texts against 2 for the equivalent numeric x. Still much better than the blank chart it replaces, but it is the numeric path, not the category one.

### Evidence

New spec `tests/unit/date-x-edge-cases.spec.js`, 5 tests. Reverting `src/` and keeping the spec, on `main`:

```
× reads the Date as its epoch on a numeric axis
    AssertionError: expected [ NaN, NaN ] to deeply equal [ 1748924002000, 1748924003000 ]
× lands on the same x as the equivalent number
    AssertionError: expected [ NaN, NaN ] to deeply equal [ 1748924002000, 1748924003000 ]
× makes isValidDate answer false rather than throw
    AssertionError: expected [Function] to not throw an error but 'TypeError: dateStr.replace is not a f…' was thrown
✓ still recognises the dates that are real
× does not take the render down
    AssertionError: expected [Function] to not throw an error but 'TypeError: dateStr.replace is not a f…' was thrown
```

The test that passes before the fix is the control: `isValidDate` still says yes to `new Date(...)`, `'2010-01-01T00:00:00.000Z'` and `'01/01/2017'`, and no to `'not a date'`. It stays green throughout, so the guard does not widen what counts as a date.

Each fix was then reverted on its own, to check that neither test is vacuous:

| reverted | result |
|---|---|
| `DateTime.js` guard only | 2 failed, 3 passed |
| `Data.js` `Date` branch only | 2 failed, 3 passed |
| neither | 5 passed |

Full unit suite: **134 files, 3181 tests, 0 failures.** `npm run lint` is clean. `npm run typecheck` reports 23 errors with and without this change — all in `Drilldown.js`, all pre-existing.

Prettier: I left the pre-existing deviations in both files alone, the way you did on #5277. `prettier --write` on these two files touches only `DateTime.js:432`, `DateTime.js:685`, `Data.js:898` and `Data.js:1775`; none of them are lines this PR adds.

### Not included

Nothing else moved. `getTimeStamp` and `parseDateWithTimezone` keep the string path they always had, and a valid `Date` never reaches the new guard because `Date.parse` succeeds before it.

### DIFF
--- a/src/modules/Data.js
+++ b/src/modules/Data.js
@@ -340,7 +340,14 @@ export default class Data {
           } else {
             this.w.axisFlags.dataFormatXNumeric = true
             this.w.axisFlags.isXNumeric = true
-            this.twoDSeriesX.push(parseFloat(x))
+            // The two lines above have already decided this x is numeric.
+            // parseFloat() reads a Date through toString(), which starts with
+            // a weekday name, so every Date arrived here as NaN. A Date's
+            // numeric value is its epoch, the same one the datetime branch
+            // above reads.
+            this.twoDSeriesX.push(
+              x instanceof Date ? x.getTime() : parseFloat(x),
+            )
           }
         }
       } else if (isXArr) {
--- a/src/utils/DateTime.js
+++ b/src/utils/DateTime.js
@@ -58,12 +58,21 @@ class DateTime {
   }
 
   /**
-   * @param {string} dateStr
+   * @param {any} dateStr - a date string, or anything isValidDate() was handed
    */
   parseDate(dateStr) {
     const parsed = Date.parse(dateStr)
     if (!isNaN(parsed)) {
       return this.getTimeStamp(dateStr)
+    }
+
+    // The salvage pass below re-spells the input and parses it again. Only a
+    // string has an alternate spelling to try: on an invalid Date, or on any
+    // other object, .replace() does not exist and the call threw. That took
+    // isValidDate() down with it, so a single bad x killed the whole render
+    // instead of being reported as "not a date".
+    if (typeof dateStr !== 'string') {
+      return NaN
     }
 
     let output = Date.parse(dateStr.replace(/-/g, '/').replace(/[a-z]+/gi, ' '))
--- a/tests/unit/date-x-edge-cases.spec.js
+++ b/tests/unit/date-x-edge-cases.spec.js
@@ -1,0 +1,105 @@
+import Data from '../../src/modules/Data'
+import DateTime from '../../src/utils/DateTime'
+import { createChart } from './utils/utils.js'
+
+// Two adjacent problems the review of #5277 named and deliberately left out of
+// that PR, because they predate it:
+//
+//   1. On a non-datetime axis a Date x went through parseFloat(), which reads
+//      the Date through its toString() ("Wed Jun 03 2026 ..."), so every point
+//      became NaN even though the surrounding code had just declared this x
+//      numeric.
+//   2. isValidDate(new Date('garbage')) threw instead of answering false,
+//      because parseDate()'s salvage pass calls .replace() on its argument and
+//      an invalid Date has no such method. That took the whole render down.
+
+const T0 = 1748924002000
+const T1 = 1748924003000
+
+function parsedX(series, xtype) {
+  const chart = createChart('line', series, xtype)
+  chart.w.globals.seriesX = []
+
+  const data = new Data(chart.w, chart)
+  const w = data.parseDataAxisCharts(series, series, chart)
+
+  return w.globals.seriesX
+}
+
+function datesAndNumbers(xtype) {
+  const [fromDates] = parsedX(
+    [
+      {
+        name: 'Date',
+        data: [
+          { x: new Date(T0), y: 1 },
+          { x: new Date(T1), y: 2 },
+        ],
+      },
+    ],
+    xtype,
+  )
+  const [fromNumbers] = parsedX(
+    [
+      {
+        name: 'Number',
+        data: [
+          { x: T0, y: 1 },
+          { x: T1, y: 2 },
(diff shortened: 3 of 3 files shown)


junedchhipa:
Verified both fixes against `b0dc6af`. Reverting each one on its own reproduces your table exactly, and the control test stays green throughout, so nothing new counts as a date; suite, lint and typecheck all match what you reported.

One wording fix for the description, not the code: an invalid `Date` doesn't reach the category path. It now looks like a plain numeric x instead, which is fine on numeric and datetime axes, but on the default category axis the chart comes out with no x axis at all. Still much better than the blank chart it replaces.

Worth mentioning too, since it's the one thing here with no test on it: the same guard un-breaks `null` and `undefined` x values, which threw from that line as well. `[{x:1},{x:null},{x:3}]` went from a blank chart to rendering correctly, a more common case than an invalid `Date`, so one assertion on it would be good insurance.

Thanks, merging.

junedchhipa (approved):

aron-intframe:
You're right on the wording, and the measurement is a bit sharper than "looks like a plain numeric x". Failing `isValidDate` means an invalid `Date` misses the `isXString || isXDate` branch entirely and lands in the final `else` in `Data.js`, which pushes `x` **unconverted** and sets both numeric flags. So `fallbackToCategory` is never set, `categoryLabels` stays empty, and the value sitting in `seriesX` is the invalid `Date` object itself, not a number. Instrumented on a category axis:

```
CLAIM1 raw pushed     = [object Date]
CLAIM1 isXNumeric     = true
CLAIM1 dataFormatXNum = true
CLAIM1 categoryLabels = []
CLAIM1 DOM xaxis label texts = 0
CONTROL numeric-x DOM xaxis label texts = 2
CONTROL numeric-x DOM xaxis label content = ["1","2"]
```

The `.apexcharts-xaxis` group is still emitted, it just carries zero label texts, and the series path draws. I've corrected that sentence in the #5279 description.

Two refinements to the axis part, from probing each type rather than reasoning about it:

```
allbad/numeric   => xlabels: ["0","1"]
allbad/datetime  => xlabels: []
allbad/category  => xlabels: []
mixed/category   => xlabels: ["1","2","3"]
mixed/datetime   => xlabels: ["00:00:00"]
```

The empty axis is not category-only: `datetime` does the same thing. Only `numeric` keeps an axis, and it does it by falling back to indices `0, 1` rather than by reading the x. And it takes a series where *every* x is bad — one invalid `Date` sitting among usable ones still labels `1, 2, 3` on a category axis.

Your `null` / `undefined` point is confirmed, and it is the bigger of the two. Reverting the guard:

```
CLAIM2 isValidDate(null)      = THREW: Cannot read properties of null (reading 'replace')
CLAIM2 isValidDate(undefined) = THREW: Cannot read properties of undefined (reading 'replace')
CLAIM2 seriesX for [1,null,3] = THREW: Cannot read properties of null (reading 'replace')
CLAIM2 DOM series paths = 0
```

With the guard in place: `seriesX = [[1, null, 3]]`, 1 series path, labels `1, 2, 3`.

Assertions are up as #5281, tests only, no `src/` change. Three of them, all red when the guard is reverted: `Tests 5 failed | 3 passed (8)` reverted against `8 passed` in place, the 5 being the 2 from #5279 plus these 3, so neither set is vacuous. Full unit suite 136 files / 3257 tests / 0 failures, `npm run lint` clean, `npm run typecheck` still the same 23 pre-existing `Drilldown.js` errors.

Not verified: all of this is jsdom, I did not open a real browser, and I did not run the e2e or interaction suites.

PR #5280, opened 2026-08-27
fix(config): keep initialConfig unadulterated when series are hidden

### LINKED ISSUE #5118: initialConfig in custom tooltip is empty when collapsedSeries are used
### Description

When new options are provided to the chart and `updateOptions` is called while there are hidden series, the series data in the `initialConfig` are cleared https://github.com/apexcharts/apexcharts.js/blob/0e7addc291d4fadbe4aa6f07d682dda106f3ebf1/src/modules/Series.js#L104

### Steps to Reproduce

1. console.log `chart.w.globals.initialConfig.series` with a hidden series
2. Update the series using `chart.updateOptions`
3. console.log `chart.w.globals.initialConfig.series` with a hidden series

### Expected Behavior
I am expecting that `initialConfig` should always provide unadulterated initial config

### Actual Behavior
`initialConfig` removes data from hidden series

### Screenshots

Image

### Reproduction Link

https://codepen.io/bmanth60/pen/ogbGpGa

aron-intframe:
Fixes #5118. The report points straight at `Series.js`, and that turned out to be the symptom rather than the cause: `globals.initialConfig` was never a snapshot at all.

### Why it emptied

`Utils.extend()` copies arrays by reference, because `isObject()` excludes them:

```js
static isObject(item) {
  return item && typeof item === 'object' && !Array.isArray(item)
}
```

So `globals.initialConfig = Utils.extend({}, config)` left `initialConfig.series === config.series`, the very same array of the very same objects. A legend collapse replaces `series[i].data` with `[]` in place, and the "initial" config lost the data along with the live one.

Fixing only that is not enough. `updateOptions()` re-captures the snapshot from `w.config`, which by then carries the collapse state, so an unrelated option change baked the emptied series back into the baseline. That capture now moves only when the call actually redefines the series; a plain `updateOptions({title})` keeps the baseline it already had.

The copy is per-series and shallow — series objects copied, data arrays shared — the same shape `globals.initialSeries` already captures. Internal edits to a series' data are property *replacements* (`series[i].data = []`, or `series[i] = 0` for non-axis charts) and neither can reach a copy made this way, so snapshotting stays O(n) instead of deep-cloning every point. That keeps the ~23ms-per-50k-update budget the lazy `initialSeries` snapshot was built to protect.

### Evidence

New spec `tests/unit/initial-config-hidden-series.spec.js`, 4 tests. On `main` (b0dc6af), with the spec and without the `src/` change:

```
✓ keeps the hidden series data after hideSeries
× keeps the hidden series data across updateOptions
    AssertionError: expected [] to deeply equal [ 4, 5, 6 ]
× does not share series objects with the live config
    AssertionError: expected [ { name: 'A', …(2) }, …(1) ] not to be [ { name: 'A', …(2) }, …(1) ] // Object.is equality
✓ still records what the chart started with
```

Each half of the fix was then reverted on its own, to confirm neither test is vacuous:

| reverted | failing test |
|---|---|
| `Globals.js` capture only | does not share series objects with the live config |
| `UpdateHelpers.js` capture only | keeps the hidden series data across updateOptions |
| neither | 4 passed |

Full unit suite: **134 files, 3180 tests, 0 failures.** `npm run lint` clean. `npm run typecheck` reports the same 23 pre-existing `Drilldown.js` errors as `main` — the two `NonNullable<...>` casts are there to keep that number from moving.

Prettier: pre-existing deviations left alone, as on #5277. `prettier --write` on the touched files moves only `Utils.js:432`, `UpdateHelpers.js:185`, `UpdateHelpers.js:304` and `UpdateHelpers.js:437`, none of which are lines this PR adds.

### Not included

`globals.initialSeries` is also refreshed from the live config during a collapse-driven re-render, so it carries the emptied series too. Nothing in this PR depends on that, `resetSeries` clears the collapsed indices before restoring, and it predates this change — mentioning it here rather than widening the diff.

### DIFF
--- a/src/modules/helpers/UpdateHelpers.js
+++ b/src/modules/helpers/UpdateHelpers.js
@@ -284,13 +284,13 @@ export default class UpdateHelpers {
       this.ctx._writeParsedAxisFlags(parsedState.axisFlags)
 
       if (overwriteInitialSeries) {
-        // initialConfig.series has aliased w.config.series since Globals.init
-        // (Utils.extend copies arrays by reference); keep that alias fresh.
+        // Refresh the initialConfig snapshot with the series the caller just
+        // defined. A copy, never the live array: this site used to assign
+        // w.config.series itself, restoring the very alias #5118 came from, so
+        // one updateSeries() would undo the capture Globals.init now makes.
         // initialSeries was already captured by parseData above through the
         // lazy-snapshot setter, so no deep clone happens here either.
         if (w.globals.initialConfig) {
-          // A copy, not the alias the comment above describes: sharing the array
-          // is what let a later collapse empty the snapshot (#5118).
           w.globals.initialConfig.series = Utils.copySeriesShallow(
             w.config.series,
           )
--- a/src/modules/helpers/UpdateHelpers.js
+++ b/src/modules/helpers/UpdateHelpers.js
@@ -161,7 +161,25 @@ export default class UpdateHelpers {
               : []
 
             // After forgetting lastAxes, we need to restore the new config in initialConfig/initialSeries
-            w.globals.initialConfig = Utils.extend({}, w.config)
+            // Two traps here. Utils.extend() copies arrays by reference, so
+            // the snapshot would hand back the LIVE series array; and
+            // w.config.series carries the legend-collapse state
+            // (`series[i].data = []` for every hidden series), so
+            // re-snapshotting it made the "initial" config forget the data of
+            // everything the user had toggled off (#5118). Only a call that
+            // actually redefines the series may move that baseline; a plain
+            // updateOptions() keeps the one it already had.
+            const prevInitialSeries =
+              w.globals.initialConfig && w.globals.initialConfig.series
+            const initialConfig =
+              /** @type {NonNullable<typeof w.globals.initialConfig>} */ (
+                Utils.extend({}, w.config)
+              )
+            initialConfig.series =
+              !options.series && prevInitialSeries
+                ? prevInitialSeries
+                : Utils.copySeriesShallow(w.config.series)
+            w.globals.initialConfig = initialConfig
             // lazy snapshot: deep clone deferred to first read
             w.globals.initialSeries = w.config.series
 
@@ -271,7 +289,11 @@ export default class UpdateHelpers {
         // initialSeries was already captured by parseData above through the
         // lazy-snapshot setter, so no deep clone happens here either.
         if (w.globals.initialConfig) {
-          w.globals.initialConfig.series = w.config.series
+          // A copy, not the alias the comment above describes: sharing the array
+          // is what let a later collapse empty the snapshot (#5118).
+          w.globals.initialConfig.series = Utils.copySeriesShallow(
+            w.config.series,
+          )
         }
         w.globals.initialSeries = w.config.series
       }
--- a/src/modules/settings/Globals.js
+++ b/src/modules/settings/Globals.js
@@ -552,9 +552,18 @@ export default class Globals {
    * Why sharing the data arrays is safe: internal "mutations" of a series'
    * data are property REPLACEMENTS (`series[i].data = []` on legend collapse),
    * which cannot reach the captured copies because each series object was
-   * copied at capture time. The one in-place mutator (appendData's push loop)
-   * re-captures immediately after mutating, so the pending snapshot never
-   * spans the mutation.
+   * copied at capture time.
+   *
+   * The one in-place mutator is appendData(), whose push loop grows the shared
+   * data array. Re-capturing after it does not undo that: the pushed points
+   * are already in the array both snapshots point at, so a snapshot taken
+   * before the append reads back as appended. That is the documented
+   * behaviour of appendData(overwriteInitialSeries = true), and the `false`
+   * case is not honoured for a separate, older reason: Data.parseData()
+   * re-captures initialSeries unconditionally on the re-render appendData
+   * triggers. Detaching would mean copying the data arrays, which is exactly
+   * the per-point cost this snapshot exists to avoid. The same exception
+   * applies to `initialConfig.series`, which is captured with the same shape.
    *
    * @param {Record<string, any>} globals
    */
@@ -573,11 +582,7 @@ export default class Globals {
         return snap
       },
       set(value) {
-        src = Array.isArray(value)
-          ? value.map((s) =>
-              s && typeof s === 'object' && !Array.isArray(s) ? { ...s } : s,
-            )
-          : value
+        src = Utils.copySeriesShallow(value)
         snap = null
         // Read-only peek for hot-path consumers (tooltip's same-length check
         // runs per update): reading it does NOT materialize the deep snapshot.
--- a/src/modules/settings/Globals.js
+++ b/src/modules/settings/Globals.js
@@ -597,7 +597,15 @@ export default class Globals {
 
     this.defineLazyInitialSeries(globals)
 
-    globals.initialConfig = Utils.extend({}, config)
+    // Utils.extend() copies arrays by reference (isObject() excludes arrays),
+    // so initialConfig.series WAS config.series. Collapsing a series replaces
+    // series[i].data with [] in place, and the snapshot lost the data with it.
+    const initialConfig =
+      /** @type {NonNullable<typeof globals.initialConfig>} */ (
+        Utils.extend({}, config)
+      )
+    initialConfig.series = Utils.copySeriesShallow(config.series)
+    globals.initialConfig = initialConfig
     globals.initialSeries = config.series
 
     globals.lastXAxis = Utils.clone(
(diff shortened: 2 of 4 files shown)


junedchhipa:
Verified against `b0dc6af`: the repro is real, your table reproduces exactly (2 failed, 2 passed), and suite, lint and typecheck all match what you reported. `npm run build` passes too, so the shared-module contract is intact; a static method on the existing default export was the right call there.

One thing to sort out before this goes in. The comment above `defineLazyInitialSeries` already documents this exact trade-off, and it names the exception: `appendData`'s push loop mutates the data array in place, and it is safe only because it re-captures `initialSeries` immediately after. Nothing re-captures `initialConfig.series`, so the new snapshot does span that mutation. After `appendData([{ data: [99] }])`, `initialConfig.series[0].data` is `[1,2,3,99]`, on `main` and on this branch alike. Not a regression, but the claim that no in-place edit can reach the copy needs narrowing, or `appendData` needs the same re-capture.

While you are in there: that setter does the identical per-series shallow copy inline, a few lines above the code you changed. Reusing `copySeriesShallow` in it would collapse the duplication and make the two snapshots provably the same shape, which the description already claims they are.

On the deferred half: `resetSeries()` after a legend hide still leaves the series empty, so the reset-home button cannot bring a hidden series back. Clearing the collapsed indices doesn't cover that, because `initialSeries` was already emptied by the collapse re-render, leaving nothing to restore from. Fine to keep out of this diff, but worth stating accurately, especially since your fix now leaves `initialConfig` holding exactly the data that path needs.

One test worth adding: `updateOptions({ series })` while a series is collapsed was also corrupt before and is clean now, and nothing covers it.

aron-intframe:
All four are right; here is what I re-ran before agreeing.

**appendData.** Reproduced, and it is not something this PR introduces: identical on `main` (`62d7661`) and on the branch.

```
P1 initialConfig.series[0].data = [1,2,3,99]
P2 initialSeries[0].data = [1,2,3,99]
P3 read-before = [1,2,3]  read-after = [1,2,3,99]
```

I tried the re-capture you offered rather than assuming it works, dropping `initialConfig.series = Utils.copySeriesShallow(w.config.series)` into `appendData` next to the `initialSeries` line:

```
WITH RECAPTURE initialConfig.series[0].data = [1,2,3,99]
```

It cannot help. The pushed points are already inside the array both snapshots point at, so re-capturing copies the series object again and lands on the same data array; only copying the data arrays would detach it, which is the per-point cost the lazy snapshot exists to avoid. So I narrowed the claim in `copySeriesShallow` and in the `defineLazyInitialSeries` note instead of adding a no-op call, and both now name appendData as the exception.

Measuring that turned up a second thing, pre-existing and equally true on `main`: `appendData(newData, false)` does not protect `initialSeries` either. P3 above read the deep snapshot *before* the append (`[1,2,3]`) and read `[1,2,3,99]` after, because `Data.parseData()` re-captures unconditionally on the re-render `appendData` triggers:

```js
    } else {
      // lazy snapshot: the globals setter stores a cheap per-series shallow
      // copy; the deep clone materializes only if something reads it
      gl.initialSeries = ser
    }
```

Out of scope here and not touched; say the word and I will file it separately.

**Dedup.** Done: the setter is now `src = Utils.copySeriesShallow(value)`. It is behaviour-neutral (reverting just that line to the old inline copy leaves the spec 7/7 green), so a test pins the shape the two capture sites now share by construction: `snapshot[0]` is not `config.series[0]`, `snapshot[0].data` is `config.series[0].data`, for both `_initialSeriesPeek` and `initialConfig.series`.

**resetSeries.** Your correction is right and my "Not included" paragraph was not. Verified:

```
P4 after hide: config= [] initialSeries= [] initialConfig= [4,5,6]
P4 after resetSeries: config= [] collapsed= []
```

Clearing the collapsed indices is not the issue; `initialSeries` was already emptied by the collapse re-render, so `resetSeries()` restores `[]` onto `[]`. `initialConfig` now holds the `[4,5,6]` that path needs, which is what makes the follow-up possible. Still out of this diff, stated correctly this time.

**The test you asked for.** Added. On `main` it fails exactly as you predicted:

```
× keeps a collapsed series intact when updateOptions redefines the series
AssertionError: expected [] to deeply equal [ 10, 11, 12 ]
```

**One gap of my own.** Reverting *only* the `_updateSeries` capture (`initialConfig.series = Utils.copySeriesShallow(...)`) left the whole suite green: that hunk had no test. It matters because assigning `w.config.series` there restores the exact alias `Globals.init` stopped making, so one `updateSeries()` would undo the fix. It now has one, and the comment above it (which still described keeping the alias fresh) is rewritten. Revert table, redone per hunk:

| reverted | failing test |
|---|---|
| `Globals.init` capture | does not share series objects with the live config; captures initialConfig.series in the same shape as initialSeries |
| `updateOptions` capture | keeps the hidden series data across updateOptions; keeps a collapsed series intact when updateOptions redefines the series |
| `_updateSeries` capture | does not re-alias the live config when updateSeries redefines it |
| setter dedup | none (behaviour-neutral by design) |
| nothing | 7 passed |

Pushed as `2851985`. `npm run unit`: 134 files, 3183 tests, 0 failures. `npm run lint` clean, `npm run typecheck` still the same 23 pre-existing `Drilldown.js` errors, `npm run build` passes. Prettier on the touched files still moves only the pre-existing spots (`Utils.js:432`, `UpdateHelpers.js:185`, `:304`, `:437`), none of them lines this PR adds.

Not verified: only the jsdom unit suite: I did not run the e2e or Playwright interaction suites, and the appendData conclusions were measured in jsdom rather than a browser.

junedchhipa:
Thanks for re-running all of this instead of taking my word for any of it. I redid the per-hunk table on my side and it comes out exactly as you have it, the behaviour-neutral setter row included. I also ran the gates against a merge with current `main` rather than your branch point, since main has moved on to 7.1.0 and has touched `UpdateHelpers` since you branched: 139 files, 3320 tests, no failures, and it merges without conflict.

You were right about the re-capture and I was wrong to suggest it. I dropped it into `appendData` myself before I got to your reply and saw `[1,2,3,99]` either way, for exactly the reason you give: the points are already inside the array both snapshots are pointing at.

One correction to the narrowed comment, though. It says detaching would mean copying the data arrays, and there is a cheaper way out: let `appendData` replace the array instead of pushing into it. With `dstSerie.data = dstSerie.data.concat(srcSerie.data)`, `initialConfig` stays on `[1,2,3]` while the live config still gets `[1,2,3,99]`, which turns the one in-place mutator into the property replacement the rest of the design already assumes. It costs one array copy per `appendData` call, and streaming callers are paying that anyway since `trimStreamingSeries` does `s.data = data.slice(...)` on every tick. Not for this PR, but worth carrying into the follow-up.

Good catch on the `_updateSeries` hunk. That one was entirely yours, and it would have quietly undone the fix on the first `updateSeries()`. And yes, please file the `parseData`/`initialSeries` one separately: that is the half users actually feel, since the reset-home button currently cannot bring a hidden series back.

One small thing for next time: those 23 typecheck errors are not all in `Drilldown.js`. It is 14 there, plus 8 in `apexcharts.js` and 1 in `Breadcrumb.js`. All drilldown-related and the set really is unchanged, so your conclusion holds either way.

Nothing blocking from me.

junedchhipa (approved):

aron-intframe:
Ran your `concat` line on today's `main` (4b4b226) before agreeing with it: replaced the push loop in `appendData` with `dstSerie.data = dstSerie.data.concat(srcSerie.data)` and re-ran the probe. It does exactly what you say, the one remaining in-place mutator becomes a property replacement and the snapshot detaches:

```
LIVE  config.series[0].data = [1,2,3,99]
SNAP  initialConfig.series[0].data = [1,2,3]
```

Stock main reads `[1,2,3,99]` in both rows. The full unit suite stays green with the change applied:

```
 Test Files  140 passed (140)
      Tests  3332 passed | 6 expected fail (3338)
```

One nuance to carry into the follow-up with it: it detaches `initialConfig` only. `initialSeries` still ends at `[1,2,3,99]` after the same call, deliberately under the default `overwriteInitialSeries = true`, and through `parseData()`'s unconditional re-capture when `false`. The comment your correction targets sits at `Globals.js:565` now, so the follow-up should rewrite that sentence along with the code.

The `parseData`/`initialSeries` half is now filed as #5283, as you asked, with jsdom measurements from current main for both symptoms: `appendData(newData, false)` not protecting the baseline, and reset-home restoring `[]` onto `[]` after a legend hide while `initialConfig.series` still holds the `[4,5,6]` that path needs. Your concat idea is noted there as an ingredient.

You're right on the typecheck split, and "same 23 pre-existing Drilldown.js errors" was sloppy shorthand on my part. Re-counted on main:

```
     14 src/modules/drilldown/Drilldown.js
      8 src/apexcharts.js
      1 src/modules/drilldown/Breadcrumb.js
```

For the record, today's main is already past your merge-gate numbers (140 files / 3338 tests, zero failures), since the heatmap-shape spec landed after your run. Not verified here: jsdom only again, no browser or e2e runs behind any of the numbers above.

PR #5281, opened 2026-08-27
test(data): cover the null and undefined x the same guard un-breaks

aron-intframe:
Follow-up to #5279, tests only, no `src/` change.

You pointed out that the guard there also un-breaks `null` and `undefined` x values, and that it was the one thing in the PR with no test on it. You were right on both counts, and it is the more common case: a gap in a data set is spelt `null` far more often than a `Date` is built from garbage.

Reverting the guard on this branch, all three new assertions go red:

```
AssertionError: expected [Function] to not throw an error but
  'TypeError: Cannot read properties of null (reading 'replace')' was thrown
AssertionError: expected [Function] to not throw an error but
  'TypeError: Cannot read properties of undefined (reading 'replace')' was thrown
```

`Tests 5 failed | 3 passed (8)` with the guard reverted, `8 passed` with it in place. The 5 are the 2 from #5279 plus these 3, so neither set is vacuous.

What the render does, measured rather than assumed. Without the guard, `[{x: 1}, {x: null}, {x: 3}]` gives 0 `.apexcharts-line` paths and 0 x axis labels. With it, 1 path and labels `1, 2, 3`.

Full unit suite: **136 files, 3257 tests, 0 failures.** `npm run lint` clean. `npm run typecheck` still 23 errors, all pre-existing in `Drilldown.js`. `prettier --write` on the spec touched only one block this PR adds, and that reformat is included.

### DIFF
--- a/tests/unit/date-x-edge-cases.spec.js
+++ b/tests/unit/date-x-edge-cases.spec.js
@@ -103,3 +103,62 @@ describe('an invalid Date as x', () => {
     ).not.toThrow()
   })
 })
+
+// The same guard un-breaks a null or undefined x, which threw from that same
+// line and is the more common way to hit it: a gap in a data set is spelt
+// `null` far more often than a Date is built from garbage. Without the guard
+// `[{x: 1}, {x: null}, {x: 3}]` renders nothing at all.
+describe('a null or undefined x', () => {
+  function dt() {
+    const chart = createChart(
+      'line',
+      [{ name: 'n', data: [{ x: 1, y: 1 }] }],
+      'numeric',
+    )
+    return new DateTime(chart.w)
+  }
+
+  it('makes isValidDate answer false rather than throw', () => {
+    const d = dt()
+
+    expect(() => d.isValidDate(null)).not.toThrow()
+    expect(() => d.isValidDate(undefined)).not.toThrow()
+    expect(d.isValidDate(null)).toBe(false)
+    expect(d.isValidDate(undefined)).toBe(false)
+  })
+
+  it('leaves the points around the gap where they were', () => {
+    const [seriesX] = parsedX(
+      [
+        {
+          name: 'gap',
+          data: [
+            { x: 1, y: 1 },
+            { x: null, y: 2 },
+            { x: 3, y: 3 },
+          ],
+        },
+      ],
+      'category',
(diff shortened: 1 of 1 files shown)


junedchhipa:
Checked this the same way: 8 passing as it stands, and 5 failing with 3 passing once I take the guard back out, which is what you reported. The three new ones are earning their place, and the suite numbers match on the nose (136 files, 3257 tests).

You are right that `null` is the case people actually hit, so it is good to have it pinned down.

One thing to be aware of rather than change: `expect(seriesX).toEqual([1, null, 3])` locks in the `null` sitting raw in `seriesX`, which is the leftover I mentioned on #5279, a non-numeric value in an array the rest of the code reads as numbers. It works well as a regression guard. Just keep in mind that if we ever tidy that up and coerce it, this is the assertion that moves.

Nothing blocking.

junedchhipa (approved):

aron-intframe:
Agreed on the leftover, and I measured what it does downstream today rather than leave it as a note. A scratch spec rendering `[{x: 1}, {x: null}, {x: 3}]` and logging the globals gives, on both axis types:

```
CATEGORY seriesX = [[1,null,3]]
CATEGORY minX = 1 maxX = 3
CATEGORY labels = [1,2,3]
NUMERIC seriesX = [[1,null,3]]
NUMERIC minX = 1 maxX = 3
NUMERIC labels = [1,2,3]
```

So the raw `null` is benign in the min/max and label paths right now: it does not get read as 0 anywhere I could observe, and `minX` stays 1 on a numeric axis too. One near-miss I should mention so nobody chases it later: in that run the numeric-axis line `d` came out as `M 0 NaN`, which looked like the `null` leaking into pixel math, but a control chart with `x: 1, 2, 3` and no `null` (and another with a `null` y) produces the same `M 0 NaN` under jsdom, where the container has zero layout dimensions. That is a test-environment artifact, not the leftover, so I am not claiming it as a finding.

And yes, `expect(seriesX).toEqual([1, null, 3])` is exactly the assertion that moves if a coercion pass ever tidies this up. That is what I want from it: it pins today's contract, and whoever changes it gets one red test pointing at the array that shifted instead of a silent behaviour change. Happy to update it alongside that work when it comes.

PR #5284, opened 2026-09-01
fix(data): stop internal re-renders from moving the initialSeries baseline

### LINKED ISSUE #5283: resetSeries() cannot restore a hidden series: parseData() re-captures initialSeries from the live config on every re-render
`globals.initialSeries` is the baseline `resetSeries()` restores, but `Data.parseData()` re-captures it from the live series on every re-render (`src/modules/Data.js:2029` on current `main`):

```js
    } else {
      // lazy snapshot: the globals setter stores a cheap per-series shallow
      // copy; the deep clone materializes only if something reads it
      gl.initialSeries = ser
    }
```

Any re-render that happens after the live series was mutated bakes the mutation into the baseline. Two user-visible consequences, both measured on current `main` (4b4b226), jsdom:

**1. The toolbar reset-home button cannot bring a hidden series back.** A legend collapse replaces `series[i].data` with `[]` in place and triggers a re-render; `parseData()` then snapshots the emptied series, so by the time `resetSeries()` runs there is nothing left to restore:

```
after hideSeries(B):  initialSeries[1].data = []  initialConfig.series[1].data = [4,5,6]
after resetSeries():  config.series[1].data = []  collapsedSeriesIndices = []
```

`resetSeries()` does clear `collapsedSeriesIndices`, but it restores `[]` onto `[]`. Since #5280, `globals.initialConfig.series` still holds the `[4,5,6]` this path needs (second column above), so a fix has an intact source of truth to restore from.

**2. `appendData(newData, false)` does not protect the baseline either.** The `overwriteInitialSeries = false` flag skips the explicit re-capture inside `appendData`, but the re-render it triggers goes through `parseData()`, which re-captures unconditionally:

```
appendData(newData, false): initialSeries[0].data = [1,2,3,99]
```

Repro for (2): line chart with `series: [{ name: 'A', data: [1, 2, 3] }]`, call `appendData([{ data: [99] }], false)`, read `w.globals.initialSeries[0].data`. Expected `[1,2,3]`, actual `[1,2,3,99]`.

Context: found while working on #5280 and filed separately at @junedchhipa's request (https://github.com/apexcharts/apexcharts.js/pull/5280#issuecomment-5480131720). One ingredient for a fix, from that same thread and verified on `main`: letting `appendData` replace the array (`dstSerie.data = dstSerie.data.concat(srcSerie.data)`) instead of pushing into it turns the one remaining in-place data mutation into a property replacement, which the snapshot design already handles.

lazerg:
Fixes #5283. The half #5280 left open: `globals.initialSeries` is what `resetSeries()` and the toolbar reset-home button restore from, but `Data.parseData()` re-captured it from the live series on every parse, so any re-render that ran after the series was mutated baked the mutation into the baseline.

A legend collapse replaces `series[i].data` with `[]` and re-renders. By the time the user hits reset-home the baseline is the emptied series, so the reset clears `collapsedSeriesIndices` and then restores `[]` onto `[]` — the hidden series never comes back.

`overwriteInitialSeries` is already the line the codebase draws between a caller redefining the input and the library handing back the series it was given, so the fix threads it into `parseData()`: `_updateSeries` passes its own flag, and `update()`'s re-render passes `false`. First render still captures, so the baseline is always set.

That alone does not cover `appendData(newData, false)`, because the push loop grows the very array the snapshot shares. Replacing the array instead — `data = data.concat(newData)`, the approach worked out in the #5280 thread — turns the last in-place mutation into a property replacement, which the snapshot design already handles. The raw-stash types (histogram, waterfall, dumbbell, streamgraph) append into their stash the same way and needed the same treatment. The `defineLazyInitialSeries` and `copySeriesShallow` notes that named the push loop as the exception are updated with it.

One more site had the same gap: `_updateOptions` re-snapshotted `initialSeries` unconditionally. The comment two lines above it already says only a call that actually redefines the series may move that baseline, which is how #5280 left `initialConfig.series`; `initialSeries` now follows the same rule.

Worth flagging as a behaviour change: a caller that holds a reference to `series[i].data` itself across an `appendData` no longer sees the new points on that array, since the property is reassigned rather than grown. Reading through `chart.w.config.series[i].data` is unaffected, and the internal consumers already reassign (`trimStreamingSeries`, `StreamScroll`).

New spec `tests/unit/initial-series-baseline.spec.js`, 7 tests. Five fail on `4b4b226` without the `src/` change (hide, reset-home, `appendData(…, false)` on both a plain and a histogram series, unrelated `updateOptions`) and two pass before and after, guarding against over-correcting: `appendData(…, true)` and a redefining `updateSeries` must still move the baseline.

Locally: 3345 unit tests pass, eslint clean, typecheck unchanged at the 23 pre-existing errors.


junedchhipa:
Direction looks right, but two things I'd fix before merge.

**1. `appendData(newData)` with the default `overwriteInitialSeries: true` no longer moves the baseline on histogram/waterfall/dumbbell/streamgraph.** That branch (`src/apexcharts.js` ~1036) appends into the raw stash and returns `this.update()`, which now parses with `false`, so nothing captures. The plain-series branch 24 lines below does honour the flag; this one needs the same capture. Net effect: `resetSeries()` after a plain `appendData()` silently drops the appended points on those four types. Previously the re-render's unconditional recapture covered this.

**2. `_updateSeries` overwrites the capture `parseData` just made** (`src/modules/helpers/UpdateHelpers.js:305`). For the raw-stash and dataReducer types, `parseData` captures the baseline from the stash (observations), then this line reassigns `w.config.series`, which at that point holds the derived rows (bin counts, accumulated pairs, downsampled window). Same story as above: it used to be healed by the next re-render, and now it sticks. Simplest fix is to drop the line and let `parseData` own the capture under the flag it was just given.

Two smaller ones: `dataReducer` charts aren't in the `derivedRaw` check, so `appendData(true)` there baselines the downsampled view rather than the true input; and the new spec covers `appendData(false)` on a raw-stash type plus `appendData(true)` only on a line chart, so the combination that regressed is exactly the untested cell.

Also worth adding `Fixes #5283` to a commit body so the issue closes and gets cross-referenced on merge.

lazerg:
All four hold up. I reproduced each one before touching anything, and 7cd38e1 fixes them.

Append path: `update()` now takes the flag and passes it to `create()`, so the parse does the capture. Both `appendData` branches hand it their own argument, which also removes the local snapshot assignment the plain branch was doing. One correction on the effect: on the four stash types the appended points do stay on screen after `resetSeries()`, because the transform re-derives from the live stash and the stash outlives the stale snapshot. The baseline was stale as you describe, but the visible loss shows up in the `dataReducer` case below.

`_updateSeries`: dropped the reassignment. `initialConfig.series` still gets its own copy there.

`dataReducer`: I kept these out of the `derivedRaw` branch. Under the reducer threshold no reduction runs and `config.series` is what draws, so an append that only fed the stash would drop the points. The plain branch now appends into both, and `parseData`'s reducer branch supplies the baseline under the flag. That is also what made an appended point vanish on the next re-render, so there is a test for it.

Spec: four new cases. Default `appendData` on a histogram, `updateSeries` on a histogram, and the two `dataReducer` ones. All four fail on 5306a78.

`Fixes #5283` is in the commit body.

Locally: 3349 unit tests pass, eslint clean, typecheck at the same 23 pre-existing errors.

junedchhipa:
All four hold up on `7cd38e1`. I reproduced each against `main`: `appendData(false)` on a histogram now leaves the baseline at the pre-append observations, the reducer stash goes 400 to 401 on append, and both the collapse-then-reset and collapse-then-`updateOptions` paths restore the hidden series instead of an empty array. Eight of the eleven new tests fail on `main`, so they are real regression tests. CI is green on the head commit (the checks summary is showing a stale run, but the check-runs API has all five passing). I also walked the internal `update()` callers, `fastUpdate`'s fallback, `refreshTokens`, `_windowResize`, treemap zoom and the Weave reserve, and every one of them wants the new `false` default.

One follow-up your fix surfaces rather than causes. With `chart.streaming` and `chart.dataReducer` both on, `main` dropped appends entirely (twenty appends and `maxX` stayed at 9). This branch fixes that, `maxX` reaches 29. But `trimStreamingSeries` only trims `config.series`, and every later parse re-reduces from the stash, so `maxPoints` no longer bounds memory for that combination: the stash grew to 30 against `maxPoints: 10`. Adding the trim next to the existing call brings it back to 10 and makes the window scroll properly, with the suite still green:

```js
trimStreamingSeries(newSeries, me.w)
if (reducerRaw) trimStreamingSeries(reducerRaw, me.w)
```

Happy to take that as a separate PR if you would rather land this one as is. Otherwise this is ready to merge. Thanks for the thorough turnaround on the earlier round.

junedchhipa:
Holding this one, and it is not anything you did. The green checks are from 1 September and `main` has moved 46 commits since, including one of mine that collides with this PR semantically rather than textually, so both GitHub's mergeable flag and the stale CI run miss it:

`2d6795801 fix(series): keep legend collapses across a data update` (11 September) touches `UpdateHelpers.js` too. It empties each collapsed row (`data: []`) on a copy of the incoming series **before** `parseData`, so a legend hide survives an `updateSeries`.

That lands right on top of this PR's central move. Once `parseData` is the single place that captures `initialSeries`, the array it captures from is the one the reconcile just emptied, so the baseline forgets the data of exactly the series the viewer toggled off. I merged your branch into current `main` and ran the suite:

```
FAIL tests/unit/initial-series-baseline.spec.js
  > still moves the baseline when updateSeries redefines the series
  AssertionError: expected [] to deeply equal [ 10, 11, 12 ]
```

That is your own control test, so merging as-is would put `main` red.

Worth saying clearly: `main` alone already fails this behaviour today. `hideSeries('B')` then `updateSeries(...)` then `resetSeries()` does not bring B back, with or without your branch. So this is not a regression you introduced, it is a guarantee this PR makes that my later commit quietly took back.

The shape that worked when I tried it: keep a reference to the caller's array before the reconcile copies it, and when a reconcile actually ran, point the baseline at that one instead of what `parseData` captured. Gating on "a reconcile ran" matters, because it leaves the raw-stash and `dataReducer` captures `parseData` owns untouched, which is the whole point of the last round. With that in place the full unit suite passes, `series-collapse-across-update.spec.js` stays green, and B is restored.

Could you rebase on `main` and reconcile with that commit? Three PRs landed today (#5300, #5304, #5308), though none of them touch your files. Everything I verified in the previous round still holds, so this is purely the base moving underneath you.

lazerg:
Rebased onto `main` and reconciled with `2d67958` in `ea0269f8`.

I keep the incoming array before the reconcile copies it. When a reconcile ran, both `initialSeries` and `initialConfig.series` come off that array rather than the parsed one. When none ran, nothing changes, so `parseData` keeps the raw-stash and `dataReducer` captures it owns.

`initialConfig.series` needed the same treatment. `Responsive.js` re-applies `initialConfig` at a breakpoint, so an emptied row there loses the data on the next resize.

Three tests cover it, all red on `1ee0abb`: the baseline after `updateSeries`, the `initialConfig` snapshot next to its `updateOptions` twin, and `hideSeries` then `updateSeries` then `resetSeries` bringing B back. Full unit suite passes at 3432, `series-collapse-across-update.spec.js` included.

PR #5288, opened 2026-09-03
fix: fix Grouped Sparkline Tooltip Synchronization

### LINKED ISSUE #5282: Group option for chart with sparkline set to true doesn't work anymore
### Description

When the sparkline mode is enabled on line charts and you use the group option. The tooltips are not synced anymore.
It also displays a black artefact on the spark charts that are not being hovered.

You can see that it is happening for spark line because it doesn't work on the demo for the spark line
https://apexcharts.com/javascript-chart-demos/dashboards/dark/

but if you use the example in this demo https://apexcharts.com/javascript-chart-demos/line-charts/syncing-charts/ 
it works fine

I am pretty sure it did work back in the days but maybe I am wrong

### Steps to Reproduce

1. go https://apexcharts.com/javascript-chart-demos/dashboards/dark/
2. hover any spark line

### Expected Behavior

when you hover a sparkline, the tooltip is displayed on all sparklines and they are synced

### Actual Behavior

1. when you hover the first spark line, the tooltip is displayed and a black artefact is displayed on the other
2. when you hover all other sparklines, the tooltip are displayed but they are not synced at all

### Screenshots

Image

Image

### Reproduction Link

I don't think it is needed since the demo is its own reproduction link

gioboa:
# New Pull Request

## Summary

Translate pointer coordinates from the hovered chart into each grouped chart's
local grid. This keeps tooltips and crosshairs synchronized when sparkline
charts appear side by side.

Previously, sibling charts received the source chart's absolute pointer
coordinates and grid element. They interpreted the pointer outside their own
plot, producing empty tooltips and edge artifacts.

No new dependencies.

Fixes #5282

## Type of change

## Testing

- Added Chromium E2E coverage for side-by-side grouped sparklines.
- Existing grouped-chart E2E still passes.
- Relevant unit tests: 74 passed.
- Build and ESLint passed.

## Checklist:

### DIFF
--- a/src/modules/tooltip/Intersect.js
+++ b/src/modules/tooltip/Intersect.js
@@ -61,8 +61,18 @@ class Intersect {
 
     if (canvasCells) {
       const seriesBound = opt.elGrid.getBoundingClientRect()
-      const clientX = e.type === 'touchmove' ? e.touches[0].clientX : e.clientX
-      const clientY = e.type === 'touchmove' ? e.touches[0].clientY : e.clientY
+      // `opt.elGrid` is THIS chart's grid, so the pointer has to be this
+      // chart's too. In a group the event belongs to whichever sibling was
+      // hovered, and `seriesHover` hands down the pointer already translated
+      // into each member's own space; pairing the raw event with a sibling's
+      // grid puts the hit test a whole chart-width off and every sibling
+      // resolves to "no cell".
+      const clientX =
+        opt.clientX ??
+        (e.type === 'touchmove' ? e.touches[0].clientX : e.clientX)
+      const clientY =
+        opt.clientY ??
+        (e.type === 'touchmove' ? e.touches[0].clientY : e.clientY)
       const hit = renderer.hitTest(
         clientX - seriesBound.left,
         clientY - seriesBound.top,
--- a/src/modules/tooltip/Position.js
+++ b/src/modules/tooltip/Position.js
@@ -301,14 +301,18 @@ export default class Position {
       if (!elGrid) return null
       const seriesBound = elGrid.getBoundingClientRect()
 
-      x = ttCtx.e.clientX - seriesBound.left
+      // Same pairing rule as the canvas hit test in Intersect: `seriesBound` is
+      // this chart's own grid, so measure from this chart's pointer. `ttCtx.e`
+      // is the raw event and in a group it belongs to the hovered sibling, so
+      // on its own it would place the box a chart-width outside the plot.
+      x = (ttCtx.clientX ?? ttCtx.e.clientX) - seriesBound.left
       if (x > w.layout.gridWidth / 2) {
         x = x - ttW
         placement = 'left'
       } else {
         placement = 'right'
       }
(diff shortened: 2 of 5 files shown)


junedchhipa:
Thanks for this, and sorry for the slow reply. The diagnosis is right and the fix does more than you claimed: hovering one panel of a trellis currently leaves most of the other panels showing empty tooltip boxes, because they were all being measured against the hovered chart's grid. Your change fixes that too, and trellis panels are implicitly grouped, so this reaches a lot further than sparkline dashboards.

Two things came out of testing it, and I have pushed both to your branch rather than sending you round the houses. Shout if you would rather own them and I will drop the commit.

**A regression.** The canvas heatmap hit test reads the raw event against `opt.elGrid`. Once that grid is the sibling's, the raw pointer is a chart-width off, the hit test finds no cell, and the sibling takes the "no hit" branch, so its tooltip is hidden entirely. Two grouped heatmaps with `chart.renderer: "canvas"` show a tooltip on the hovered chart only. It needs the canvas renderer feature bundled, which is why nothing in the suite caught it.

**A half-fix.** `followCursor` placement anchors the box to the pointer and has the same mismatch. It was already wrong before your branch, but with the content fixed it trades an empty box for a populated one sitting outside the sibling's plot, which reads worse. Fixing it alongside seemed better than shipping that.

Both are the same rule: wherever a coordinate is measured against a chart's own grid, it has to come from that chart's translated pointer. I also added a spec for the two cases, since neither was reachable from the existing suite.

Full interaction suite is green at 571, and all 313 sample snapshots are unchanged.

Two smaller things I left for you:

- `opt.elGrid.getBoundingClientRect()` and the pointer are the same for every iteration but are recomputed inside the `forEach`. On a sixteen-panel trellis that is 32 forced layouts per mouse move instead of 17. Worth hoisting above the loop.
- Swapping `hoverArea` to each chart's own paper is defensible but unrelated to this fix, and `hoverArea` is only read for the zoom/pan cursor classes. Either keep it with a line in the commit body saying why, or drop it.

Once those are settled this is good to go, and I would like it early in the 7.2 cycle so it gets a full RC of soak rather than landing just before the release.

gioboa (commented):
Thanks @junedchhipa I addressed you suggestions

junedchhipa:
Both changes are exactly right, thanks. Dropping the `hoverArea` swap in particular: it is read only for the zoom/pan cursor classes, and that code already evaluates the zoom/pan state against the target chart, so keeping the swap would have left a path where a sibling with zoom turned off could strip the cursor class off the chart the pointer is actually over. Restoring the original pairing avoids that entirely.

Verified on the branch and again with current main merged in, since a few things landed today: interaction suite green at 577, unit at 3346, and all 313 sample snapshots unchanged. Merges cleanly, no conflicts.

One housekeeping note: I tidied the last commit subject (it read `handling2`, and the change is a perf fix rather than a chore) and force-pushed the message-only amend, so please re-pull before any further work on this branch. Your authorship is untouched.

Merging. Thanks for the fix and for turning the follow-ups around so quickly.

PR #5289, opened 2026-09-05
feat(locales): add Azerbaijani (az) locale

jamalkamaladdin:
Adds an Azerbaijani (`az`) locale following the same structure as `src/locales/en.json`: full and short month names, full and short day names, and toolbar labels (export, menu, selection, zoom, pan, reset, measure).

No other files were changed — `dist/locales/` is rebuilt from `src/locales/` at release time (not per-PR), matching the pattern of the last single-locale addition (Galician, #5096).

**Verification**

- Key structure identical to `en.json` — 0 missing, 0 extra, same order.
- 12 / 12 month names, 12 / 12 short months, 7 / 7 day names, 7 / 7 short days.
- Month and day names use standard Azerbaijani orthography, Latin script.

### DIFF
--- a/src/locales/az.json
+++ b/src/locales/az.json
@@ -1,0 +1,56 @@
+{
+  "name": "az",
+  "options": {
+    "months": [
+      "Yanvar",
+      "Fevral",
+      "Mart",
+      "Aprel",
+      "May",
+      "İyun",
+      "İyul",
+      "Avqust",
+      "Sentyabr",
+      "Oktyabr",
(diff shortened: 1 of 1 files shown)


junedchhipa:
Thanks! Checked the key structure against `en.json` and it matches exactly, and the month and day names read correctly.

You are right that no `dist/` change is needed: only `en.json` is bundled, and everything else in `src/locales` is copied across at build time, so `az.json` will ship with the next release build.

Merging.

PR #5290, opened 2026-09-05
fix(tooltip): keep heatmap and treemap tooltips inside the plot area

### LINKED ISSUE #5121: Tooltip exceeding screen limits in mobile screen

jamalkamaladdin:
Fixes #5121

## Problem

A treemap tooltip is placed beside the hovered cell by the legacy path in
`Intersect.handleHeatTreeTooltip`. That path offsets the box by its own width:

```js
x = cx + ttWidth / 2 + width
if (x > gridWidth / 2) {
  x = cx - ttWidth / 2 + width
}
```

When the label is long and the chart is narrow, `x` goes negative and the box
hangs off the left edge, so most of the text is unreadable. The arrow-mode
heatmap path a few lines above already clamps its placement into the plot area
(`finalX` bounds check), the legacy path does not.

## Measurement

Six-cell treemap, 350px chart in a 375px viewport, company names as labels.
Tooltip left edge relative to the viewport, before the fix:

| cell | tooltip left | tooltip right | result |
|---|---|---|---|
| 0 | 0 | 202 | inside |
| 1 | 2 | 200 | inside |
| 2 | -254 | 2 | off-screen by 254px |
| 3 | 2 | 230 | inside |
| 4 | -234 | 2 | off-screen by 234px |
| 5 | -226 | 2 | off-screen by 226px |

After the fix all six sit inside the plot area (left edge at the grid's left
bound, right edge well inside the viewport).

## Fix

Clamp the legacy placement horizontally at the end of the function, using the
same grid bounds the arrow-mode path uses. The clamp runs after the
`followCursor` branch so it covers that mode too.

Vertical placement is untouched: no vertical overflow showed up in the
measurement, so this stays a minimal change.

## Test

`tests/interaction/specs/treemap-tooltip-bounds.spec.js` hovers every cell of a
treemap in a 420px viewport with labels long enough to overflow, and asserts the
tooltip stays inside the plot area and on screen.

Verified both ways: the test fails on `main` without the source change and
passes with it.

### DIFF
--- a/src/modules/tooltip/Intersect.js
+++ b/src/modules/tooltip/Intersect.js
@@ -198,6 +198,22 @@ class Intersect {
         (y > w.layout.gridHeight / 2 ? ttCtx.tooltipRect.ttHeight : 0)
     }
 
+    // A label wider than the space beside the cell pushes the box out of the
+    // plot area — on a narrow screen it then hangs off the viewport and the
+    // text is unreadable. Clamp it back into the plot area horizontally, the
+    // same way the arrow-mode path above does.
+    const ttWidth = ttCtx.tooltipRect.ttWidth || 0
+    const elGridRect = opt.elGrid
+      ? opt.elGrid.getBoundingClientRect()
+      : null
+    const gridLeft = elGridRect
+      ? elGridRect.left - w.dom.elWrap.getBoundingClientRect().left
+      : w.layout.translateX
+    const gridRight = gridLeft + w.layout.gridWidth
+
+    if (x + ttWidth > gridRight) x = gridRight - ttWidth
+    if (x < gridLeft) x = gridLeft
+
     return {
       x,
       y,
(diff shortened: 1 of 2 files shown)


junedchhipa:
Verified against current main, including the case where this and the grouped-tooltip fix both act on the same function: a grouped treemap now keeps every tooltip flush inside its own panel. Suites green, all 313 sample snapshots unchanged.

I retitled this to mention heatmaps, since the clamp also moves them when `tooltip.arrow` is off, and anyone reading the changelog should see that.

Merging, thanks.

PR #5291, opened 2026-09-05
fix(selection): keep resize handles inside the data range

### LINKED ISSUE #5123: Brush selection doesn't work properly - fails to calculate correct position of zoomPanSelection
### Description

The selection handles in the Brush Chart do not respect the data boundaries when resizing the selection area. When a user drags a selection handle beyond the minimum or maximum data limit, the selection continues to expand indefinitely into empty space, rather than being constrained to the actual data range.

### Steps to Reproduce

1. Navigate to the official [Brush Chart demo](https://apexcharts.com/javascript-chart-demos/line-charts/brush-chart/)
2. Attempt to drag either selection handle beyond the existing data boundaries
3. Observe that the selection area extends indefinitely beyond valid data ranges

### Expected Behavior

The selection area should be constrained to the minimum and maximum values of the dataset. Selection handles should stop at the chart's data boundaries and not allow extension into areas without data.

### Actual Behavior

The selection handles can be dragged indefinitely beyond the data limits, creating an invalid selection range that doesn't correspond to actual data points.

### Videos

[*(Please attach any screenshots or screen recordings that demonstrate the issue)*](https://github.com/user-attachments/assets/2a8c5718-8fc1-4bf2-bc07-f285d4df2234)

### Environment
- ApexCharts Version: [as seen on demo page]
- Browser: Chrome Version 141.0.7390.108 (Official Build) (64-bit)
- OS: Windows 11

jamalkamaladdin:
Fixes #5123

## Problem

Dragging a selection handle past the first or last data point keeps expanding
the selection into empty space. The target chart of a brush pair follows it and
scrolls into a range where nothing is drawn.

A body drag of the selection was always bounded by `this.constraints`
(`ZoomPanSelection.js:82`). A handle resize never was.

## Measurement

`samples/vanilla-js/line/brush-charts.html`, gridWidth 585.5px, data from
2024-09-30 to 2025-03-28. Handle dragged 400px past the plot:

| handle | before | after |
|---|---|---|
| right, pixels past the grid | 246.3px | 0.0px |
| right, days past the data max | 75.3 days | 0.0 days |
| left, rect x | -86.0px | 0.0px |
| left, days before the data min | 26.3 days | 0.0 days |

After the fix the target chart's max lands exactly on 2025-03-28.

## Root cause and where the fix sits

The unbounded write is in `src/svg/SVGSelectable.js:148-167`: `onMove` sets the
rect's `x` and `width` straight from the pointer, its only guard being
`if (newW < 0) newW = 0`. The consumer, the `resize` listener at
`ZoomPanSelection.js:525`, then reads those attributes into
`w.interact.selection` and the emitted range.

The clamp is applied in `ZoomPanSelection.js` rather than in `SVGSelectable`
on purpose: `SVGSelectable` is a generic SVG helper with no knowledge of the
plot area, while the grid bounds belong to the chart module. Happy to move it
if you prefer the guard at the source.

The clamp rewrites the rect itself, not just the reported numbers, because
`selection-geometry.spec.js` asserts that the reported range matches the range
implied by the rect pixels.

## Scope

Every handle resize of the persistent selection rect, so brush charts and the
selection toolbar mode alike. Both use the same rect and the same handles, and
the clamp target `0..gridWidth` is by definition `minX..maxX` under
`AxisMapping`, which is the box body drags already obeyed.

Drag to zoom is untouched: `selectionDrawing` bounds its own rect to the grid
before this code runs. Nothing that is actually drawn becomes unselectable.

One case is worth flagging: on a chart that sets `xaxis.min` or `xaxis.max`
wider than its data, the clamp binds to the visible domain rather than to the
data points. The selection still stays on the chart and everything drawn stays
selectable, but the two boundaries only coincide when the domain is data
driven, as it is on the brush demo.

## Test

`tests/interaction/specs/brush-selection-bounds.spec.js`, four tests. With the
fix all four pass. Without it (verified by stashing only the source file and
rebuilding) three fail, the failing value being 246.27px, which matches the
manual measurement. The fourth passes either way by design: it guards that an
in-bounds resize is not frozen by the clamp.

`zoom-pan-selection.spec.js` and `selection-geometry.spec.js` stay green,
including the resize handle and full width brush cases.

### DIFF
--- a/src/modules/ZoomPanSelection.js
+++ b/src/modules/ZoomPanSelection.js
@@ -21,6 +21,11 @@ const FRAME_MS_60FPS = 16.6667 // one frame at 60fps; normalizes friction to the
 // A pan nudge shifts the x-window by this fraction of the grid width.
 const PAN_NUDGE_DIVISOR = 15
 
+// Left edge of the x-domain in the plot-origin pixel space the selection rect
+// lives in. Under AxisMapping, minX maps to this pixel and maxX maps to
+// gridWidth, so the pair is the data boundary expressed in rect coordinates.
+const PLOT_ORIGIN_PX = 0
+
 /**
  * ApexCharts Zoom Class for handling zooming and panning on axes based charts.
  *
@@ -523,6 +528,11 @@ export default class ZoomPanSelection extends Toolbar {
         })
         .resize()
         .on('resize', () => {
+          // SVGSelectable moves a handle freely (its only guard is a negative
+          // width), so pull the rect back inside the x-domain before anything
+          // reads it. Without this a handle dragged past the first or last point
+          // keeps widening the rect into empty space (#5123).
+          this._clampSelectionRectToPlot()
           if (w.interact.selectionEnabled) {
             // A handle resize re-reports the selection through the SAME shared
             // mapping (and debounce) as the rect-body drag. The old path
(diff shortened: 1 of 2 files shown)


junedchhipa:
Thanks, and this is the right place for the guard. The SVG helper has no idea where the plot area is, so the chart module has to own that bound.

I confirmed it covers the selection toolbar mode as well as brush charts, and that a drag that stays in range still follows the pointer exactly rather than getting stuck.

Worth noting for anyone who lands here later: if the axis min/max is set wider than the data, the handle stops at the edge of the visible axis rather than at the last point. That is the sensible behaviour, and everything drawn stays selectable.

Merging.

PR #5292, opened 2026-09-06
fix(annotations): keep long point annotation labels inside the grid

### LINKED ISSUE #5106: Point annotation labels are cut of

jamalkamaladdin:
Fixes #5106

## Problem

A point annotation's label is drawn at the point's x position with no width
check, and the default `label.textAnchor` is `middle`. Near the left or right
edge of the plot, half of a long label falls outside the chart's SVG viewport,
which has `overflow: hidden`, so the text is clipped instead of being moved
inward.

## Measurement

Line chart with categories, one long label on the first x position and one on
the last. Overflow past the chart edge, before the fix:

| annotation | overflow |
|---|---|
| first point (left edge) | 66.6px clipped |
| last point (right edge) | 105.3px clipped |

After the fix both labels sit flush inside the grid, with no clipping.

## Fix

`getConstrainedLabelX()` measures the label with `Graphics.getTextRects`, the
same zoom safe helper `getTextBasedOnMaxWidth` already uses, works out the
label's left and right edge for the active `textAnchor` (`start`, `end` or the
default `middle`), and shifts x inward by exactly the overflow when an edge
falls outside `0..gridWidth`. That range is the same local coordinate box the
marker's own x is already clamped into by `Helpers.getX1X2`.

Labels that fit are untouched, so nothing moves on charts that were already
fine.

## Test

`tests/interaction/specs/point-annotation-bounds.spec.js` adds two long point
annotations at the first and last category through the runtime
`chart.addPointAnnotation` API and asserts both label boxes stay inside
`.apexcharts-grid`.

Verified both ways: the test fails on `main` without the source change
(`label.left` 239.4 against a required 365.5) and passes with it.

## Note

@asmenezes reported having a fix for this in the issue thread back in
September 2025 but was blocked on running the test suite and on branch
permissions. If that work is still coming, this PR can be closed in its favour.

### DIFF
--- a/src/modules/annotations/PointsAnnotations.js
+++ b/src/modules/annotations/PointsAnnotations.js
@@ -170,6 +170,12 @@ export default class PointAnnotations {
    * "moved into the chart" behaviour users expect. Nudge the label's x
    * inward just enough to keep its full rendered width inside the grid.
    *
+   * What has to fit is the label's BOX, not its text node: `label.style.background`
+   * is set by default, and `Helpers.annotationsBackground` draws that background
+   * from the rendered text's bounds plus `label.style.padding`. Clamping the text
+   * alone leaves the drawn box overhanging by the padding, which still clips on a
+   * chart whose grid meets the SVG edge (a sparkline, or zero chart padding).
+   *
    * @param {string} text
    * @param {number} x anchor x, already including `label.offsetX`
    * @param {Record<string, any>} label `anno.label`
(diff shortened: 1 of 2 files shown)


junedchhipa:
Merging this over #5276, which fixed the same issue a few weeks earlier: shifting by the overflow leaves labels that already fit untouched, and measuring at the rendered font weight matters for bold labels. Credit noted on that PR.

One thing I pushed to your branch: the visible label is the background rect, not the text node, and that rect is the text bounds plus `label.style.padding`. Clamping the text alone still left the drawn box 5px outside the grid at each edge, which on a chart whose grid meets the SVG edge is the original symptom. The commit folds the padding into the measured edges and moves the assertion onto the rect. Verified both ways, and the box now lands exactly on the grid bounds.

Merging, thanks for the clear write-up.

PR #5293, opened 2026-09-06
fix(bar): stack combo columns on the layer below, not the axis

### LINKED ISSUE #5105: The stacked property does not work correctly in mixed, line, and column charts.
### Description
In version 3.47.0, the mixed column and line chart works correctly, but when migrating to version 4.7.0, the stacked property does not seem to work correctly. 

### Expected Behavior

Image

### Actual Behavior

Image

### Reproduction Link
https://codesandbox.io/p/sandbox/linea56-wojcot?file=%2Fsrc%2Findex.js%3A3%2C3-2455%2C5

jamalkamaladdin:
Fixes #5105

## Problem

In a mixed chart with `chart.stacked: true`, columns declared after a line or
area series restart from the axis instead of stacking on the column below.

The order matters, which is why the existing samples never caught it: with the
line declared last the stack is correct today. Declaring the line first or in
the middle flattens every column that follows it.

## Measurement

Three column series plus one line series, categories A to D, animations off,
grid baseline at y=308.4. Column rect bottom edge per series, line declared
first:

| build | ColA bottom | ColB bottom | ColC bottom |
|---|---|---|---|
| 3.47.0 | 314.4 (axis) | 262.0, top of ColA | 235.7, top of ColB |
| 4.7.0 | 308.4 (axis) | 308.4 (axis) | 308.3 (axis) |
| main, before | 308.4 (axis) | 308.4 (axis) | 308.3 (axis) |
| main, after | 308.4 (axis) | 257.0, top of ColA | 231.2, top of ColB |

With the line in the middle the same fix gives ColB bottom 257.0 and ColC
bottom 231.3, both correct.

## Root cause

`src/charts/BarStacked.js`, `drawStackedColumnPaths` and `drawStackedBarPaths`.

`w.labelData.seriesGroups` buckets every series of a group, lines and areas
included (`src/modules/Data.js:821-827`). `prevY`, `prevYF` and `prevYVal` are
pushed once per series this renderer actually draws
(`BarStacked.js:277-282`). Taking the stack position from
`seriesGroup.indexOf(...)` therefore counted non bar members too, so a column
declared after a line indexed past the end of `prevY`, `bYP` came back
undefined, fell through to `w.layout.gridHeight` and the bar restarted at the
axis.

On 3.47.0 the same line read `seriesGroup.indexOf(w.config.series[i].name)`
with `i` being the bar loop index, which always resolved back to `i`, so it was
accidentally correct. The 4.x rewrite switched it to the global series index,
which is where this starts.

## Fix

Take the position from `this.groupCtx.prevY.length` (and `prevX.length` for
the horizontal path), which counts the bars this renderer has drawn for the
group by construction. No behaviour change outside a mixed group.

## Test

`tests/interaction/specs/mixed-stacked-columns.spec.js`, two cases: line
declared before the columns, and line declared after as a control.

Without the source change the first case fails
(`ColB should start at the top of ColA in category A, expected < 1, received
51.39`) and the control passes, which is the point. With the change both pass.

## Regression checks

Full interaction suite run three times: 570 passed, 0 failed in each run, the
four `stacked-*` specs green (33/33) in each. Unit suite 3332 passed across
140 files.

Beyond the suite, all 30 samples containing `stacked: true` were rendered
before and after and their `.apexcharts-bar-area` `d` attributes compared: 30
of 30 byte identical, so this is a no-op for every non mixed stacked chart,
including the grouped and border radius ones.

## One thing left alone

The numeric x branch of the same condition (`BarStacked.js:582-585`) compares
`w.seriesData.seriesX[realIndex - 1][j]` against `seriesX[realIndex][j]`. In a
combo chart `realIndex - 1` can be a line series rather than the previous bar,
which looks like the same bug class. It was equally wrong before this change,
and the reproduction here uses a category axis so it never reaches that branch.
Left untouched rather than fixed blind. Happy to follow up separately.

### DIFF
--- a/src/charts/BarStacked.js
+++ b/src/charts/BarStacked.js
@@ -446,12 +446,10 @@ class BarStacked extends Bar {
       prevBarW = prevBarW + this.groupCtx.prevXF[k][j]
     }
 
-    let gsi = i // an index to keep track of the series inside a group
-    if (/** @type {Record<string,any>} */ (w.config.series[realIndex]).name) {
-      gsi = seriesGroup.indexOf(
-        /** @type {Record<string,any>} */ (w.config.series[realIndex]).name,
-      )
-    }
+    // Where this series sits in its group's stack. See the note in
+    // drawStackedColumnPaths: the index has to count the series this renderer
+    // has already drawn for the group, which is exactly what prevX holds.
+    const gsi = this.groupCtx.prevX.length
 
     if (gsi > 0) {
       let bXP = zeroW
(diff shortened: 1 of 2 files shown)


junedchhipa:
Excellent write-up, and the diagnosis holds up. I re-rendered every sample and the stacked ones are unchanged, so this is a no-op outside mixed charts, and it turns out to fix the numeric x-axis case you flagged as untouched.

Two related problems survive on the same chart, both of which predate this change: the rounded corner ends up on the wrong end of the stack, and a 100% stacked combo throws outright when the line is declared first. I will track those separately rather than hold this up.

Merging.

PR #5298, opened 2026-09-06
fix(bar): keep the stack border-radius off series the bar renderer never draws

### LINKED ISSUE #5296: Stacked combo chart: the rounded corner lands on the wrong bar
In a stacked combo chart, `plotOptions.bar.borderRadius` rounds the wrong end of the stack. Whichever end of the series array the line (or area) series sits at, that end of the column stack is left square, because the non-bar series claims the cap.

## Reproduction

```js
new ApexCharts(el, {
  series: [
    { name: "Trend", type: "line",   data: [180, 290, 400, 510] },
    { name: "ColA",  type: "column", data: [100, 200, 300, 400] },
    { name: "ColB",  type: "column", data: [50, 60, 70, 80] },
    { name: "ColC",  type: "column", data: [30, 30, 30, 30] },
  ],
  chart: { type: "line", height: 400, stacked: true, animations: { enabled: false } },
  plotOptions: { bar: { borderRadius: 8 } },
  stroke: { width: [3, 0, 0, 0] },
  xaxis: { categories: ["A", "B", "C", "D"] },
})
```

**Expected:** ColA (bottom of the stack) is rounded at the bottom and ColC (top) is rounded at the top.

**Actual:** ColC is rounded at the top, ColA is square. Move the line to the end of the array and it inverts: ColA is rounded at the bottom and ColC is square.

## Cause

`createBorderRadiusArr` resolves the outermost segment of each stack from `getStackedSeriesIndices`, which buckets by global series index across every series in the group. Line and area series land in those buckets even though the stacked bar renderer never draws them, so the first or last index in the bucket can be a series that has no bar, and the cap is assigned to nothing.

Both are in `src/charts/common/bar/Helpers.js`.

## Notes

This is the same root cause family as #5105 (a group bucket holding series the bar renderer does not draw) but a different code path, and it was not addressed by that fix. It reproduces identically before and after it, on 7.2.0-rc.1.

lazerg:
In a stacked combo chart, `plotOptions.bar.borderRadius` rounds the wrong end of the stack: whichever end of the series array a line or area series sits at, that end of the column stack stays square, because the corner gets assigned to a series with no bar.

`createBorderRadiusArr` resolves the outermost segment of each stack from `getStackedSeriesIndices`, which buckets every series in a group by its global index, line and area included. Those non-bar series still landed in the bucket, so the first or last index in it could be a series the bar renderer never draws, and the cap it picked never rendered.

`getStackedSeriesIndices` now filters each bucket against `w.globals.columnSeries.i`, the list the combo chart already builds of which series it actually draws as bars. A line or area series can no longer take either end of the stack.

Same root cause family as #5105, a different code path, both in `src/charts/common/bar/Helpers.js`.

Fixes #5296

### DIFF
--- a/src/charts/common/bar/Helpers.js
+++ b/src/charts/common/bar/Helpers.js
@@ -506,13 +506,26 @@ export default class Helpers {
    * not grouped. Order within a bucket follows series order, which is stacking
    * order.
    *
+   * `w.globals.columnSeries` (when set) is the combo chart's own list of which
+   * series it draws as bars; a line or area series never occupies a stack
+   * segment, so it is filtered out here rather than left to compete for the
+   * outermost slot (#5296).
+   *
    * @param {number} numSeries
    * @returns {number[][]}
    */
   getStackedSeriesIndices(numSeries) {
-    const groups = this.w.labelData.seriesGroups
+    const w = this.w
+    const groups = w.labelData.seriesGroups
+    const barIndices = w.globals?.columnSeries
+      ? new Set(/** @type {any} */ (w.globals.columnSeries).i)
+      : null
+    const isBar = (/** @type {number} */ i) => !barIndices || barIndices.has(i)
+
     if (!groups || groups.length < 2) {
-      return [Array.from({ length: numSeries }, (_, i) => i)]
+      return [
+        Array.from({ length: numSeries }, (_, i) => i).filter(isBar),
+      ]
     }
 
     /** @type {number[][]} */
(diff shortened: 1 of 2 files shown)


lazerg:
This failure predates my change. The same iot-device-health screenshot flake failed directly on main earlier today (run 34018745623). My diff does not touch trellis or line rendering at all.

junedchhipa:
Thanks, and that was a quick turnaround. Filtering the buckets against the list the combo chart already keeps of what it draws as bars is the right lever, and handling the unset case so non-combo charts keep their old path is exactly the trap I was worried about.

Verified here: both series orders now cap the real ends of the stack, and plain stacked, grouped stacked and mixed-sign charts are all unchanged, with no movement in any of the sample snapshots.

Merging.

PR #5300, opened 2026-09-07
fix(ssr): estimate text bounds so the DOM shim can measure labels

### LINKED ISSUE #5299: SSR: horizontal bar data labels are vertically misaligned without text bounds
### Summary

`apexcharts/ssr` positions horizontal-bar data labels visibly above the bar centre, while the same options rendered with the browser entry point are vertically centred.

### Reproduction

```js
const apexChartsModule = require("apexcharts/ssr")
const ApexCharts = apexChartsModule.default || apexChartsModule

const svg = await ApexCharts.renderToString({
  chart: { type: "bar", width: 600, height: 180, animations: { enabled: false } },
  series: [{ name: "Count", data: [4, 2] }],
  xaxis: { categories: ["First", "Second"] },
  plotOptions: { bar: { horizontal: true, dataLabels: { position: "center" } } },
  dataLabels: { enabled: true, style: { fontSize: "12px" } },
}, { width: 600, height: 180 })
```

Render the resulting SVG to an image. The `4` and `2` labels sit at the top of their bars rather than being visually centred. Rendering the same options with `apexcharts/client` in a browser centres them correctly.

### Suspected cause

The SSR DOM shim does not provide `getBBox()`. `SVGElement#bbox()` therefore falls back to `{ x: 0, y: 0, width: 0, height: 0 }`, while the horizontal bar data-label calculation uses `textRects.height / 2`. The browser renderer has real SVG text bounds, so the two paths get different Y coordinates.

### Expected behaviour

`renderToString` should produce horizontally oriented bar data labels with the same vertical alignment as the browser renderer, including charts with many categories.

### Versions

Observed in `apexcharts@5.16.0`. I inspected `7.1.0` and the SSR fallback plus the relevant data-label calculation are still present.

lazerg:
`apexcharts/ssr` puts horizontal bar data labels above the bar centre. The same options rendered in a browser centre them.

The SSR DOM shim has no `getBBox()`, so `SVGElement#bbox()` falls back to a zero rect and every `Graphics.getTextRects()` call in SSR reports a 0x0 label. The horizontal bar label offset adds `textRects.height / 2`, which becomes 0, and the x-axis and y-axis label sizes that set the plot area collapse the same way.

`SSRElement` now estimates the box for `text` and `tspan` nodes from the font size and the content, using average sans-serif metrics. Other elements keep returning an empty rect. On the reported chart the label baseline lands at y=32.087 against a bar centre of 28.085; Chrome gives 32.087 and 28.087 for the same chart.

Fixes #5299

### DIFF
--- a/src/ssr/DOMShim.js
+++ b/src/ssr/DOMShim.js
@@ -4,6 +4,14 @@
  * Provides just enough SVG element emulation to create chart structure without full DOM
  */
 
+// Approximate sans-serif metrics, in em, used to estimate SVG text bounds when
+// no font engine is available to measure the glyphs.
+const CHAR_WIDTH_EM = 0.55
+const ASCENT_EM = 0.917
+const DESCENT_EM = 0.25
+const LINE_HEIGHT_EM = 1.1
+const DEFAULT_FONT_SIZE = 11
+
 /**
  * Mock SVG element for SSR environment
  */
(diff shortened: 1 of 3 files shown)


junedchhipa:
Verified the fix. The label baseline lands at 32.087 against a bar centre of 28.085, matching what you reported, and the x-axis height reserve now responds (grid height 135 to 112 on that chart). Good call taking `LINE_HEIGHT_EM` from `TspanWrapper.newLine()`'s `dy: 1.1em` rather than guessing it. I also checked that giving every element a `getBBox` is safe: the shim's `querySelector` and `querySelectorAll` return `null` and `[]`, so the label-removal and datalabel-background paths that call it directly never run in SSR at all.

The gap is that only half the measurement flows through. `Dimensions/YAxis` (and three other sites) call `getTextRects(..., useBBox: false)`, which reads `getBoundingClientRect()`, and `SSRElement.getBoundingClientRect()` still returns `_ssrWidth` / `_ssrHeight`, i.e. zero for a virtual text node. So the y-axis label *width* reserve is unchanged by this PR: a chart with y labels `1, 2, 3` and one with `2,500,000` both emit `translate(27, 30)`. That is the measurement that shifts the whole plot horizontally, so it is arguably the bigger half of the family #5299 belongs to.

Delegating `getBoundingClientRect()` to the same estimate for `text` and `tspan` gives 45.15 and 69.35 for those two charts, with the unit suite still green. One caveat if you take it on: `getTextRects` passes a `rotate(...)` transform and `getBBox` is pre-transform, so a rotated y-axis title would reserve its text *length* rather than its height. I measured `translateX` at 149 where roughly 85 is realistic. Worth handling in the same change rather than trading one layout error for another.

Two minor notes, worth a comment if nothing else: `parseFloat('1rem')` yields a 1px font, and `font-weight` is ignored, so bold labels under-measure.

PR #5304, opened 2026-09-09
fix(xaxis): honor inferred numeric tick semantics

### LINKED ISSUE #5086: x axis ticks are misaligned in bar charts that have more than 10 ticks

lovasoa:
## What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Refactor
- [ ] Other (please describe)

## What is the current behavior?

Numeric XY parsing sets `isXNumeric`/`dataFormatXNumeric`, but an omitted `xaxis.type` remains configured as `category` for vertical bar charts. That splits the axis behavior: `Range` skips the under-30 numeric tick heuristic, while `Formatters` rounds generated fractional coordinates as category labels. In the complete two-series `x = 1..12` reproduction this places integer-looking labels between the matching bar groups.

Fixes https://github.com/apexcharts/apexcharts.js/issues/5086.

## What is the new behavior?

Inferred numeric XY axes use the existing effective semantics of explicitly numeric axes: small series select one interval per point, and default labels retain the precision needed to describe fractional scale coordinates. Explicit `tickAmount`, datetime axes, converted categories, string categories, horizontal bars, and user formatters keep their existing paths and precedence.

This is an alternative that supersedes, rather than stacks on, @waterWang's https://github.com/apexcharts/apexcharts.js/pull/5259. Credit to that PR for identifying the missing small-series Range behavior and adding the first focused tests. Its Range-only `dataPoints - 1` change fixes consecutive integers under 30 points, but cannot by itself make labels truthful for irregular scales (for example, `1, 2, 10` produces a `5.5` coordinate) or scales with 30+ points, where the width-based tick count can still produce fractional coordinates. Extending inferred numeric semantics through the default formatter covers those cases without forcing every small bar chart to be numeric.

## What does this PR do?

- Extends the existing small numeric-scale tick rule to parser-confirmed numeric XY data when it is not datetime, converted-category, or horizontal-bar data.
- Routes the same inferred numeric data through the existing precision-preserving default numeric formatter.
- Adds deterministic rendered-label and scale tests for two complete `1..12` series, irregular data, a 30-point fractional scale, explicit `tickAmount`, and a user formatter.

### Before

Before: fractional tick coordinates are displayed as misleading integers between the matching bar groups

### After

After: integer x-axis labels align with the corresponding complete bar groups

Both screenshots use a fixed-width bar chart, omit `xaxis.type` and `xaxis.tickAmount`, enable data labels, and provide two complete series with `x = 1..12` and `y = x`.

## Validation

- Upstream CI: lint, Node 20/22 Test & Build, and Node 20/22 Test Reproducibility all passed.
- `npx vitest run tests/unit/xaxis.spec.js tests/unit/formatters.spec.js` (91 passed)
- `yarn unit` (3,351 passed, 6 expected failures)
- `yarn lint` (passed)
- `yarn build` (passed)
- E2E sample runner completed all 313 samples under local Chromium; 20 unrelated image/canvas snapshot comparisons differed from repository baselines. No failed sample exercised this numeric-axis change.
- `yarn typecheck` currently fails on pre-existing drilldown typing errors on `main`; none are in touched files.

## Related history

- https://github.com/apexcharts/apexcharts.js/issues/4858
- Final unresolved #4858 report: https://github.com/apexcharts/apexcharts.js/issues/4858#issuecomment-2557902376
- Related line-chart issue: https://github.com/apexcharts/apexcharts.js/issues/5015
- Related merged Range fix: https://github.com/apexcharts/apexcharts.js/pull/5045
- Corrected complete-series SQLPage reproduction: https://github.com/sqlpage/SQLPage/issues/733#issuecomment-5601787175

### DIFF
--- a/src/modules/Formatters.js
+++ b/src/modules/Formatters.js
@@ -107,9 +107,14 @@ class Formatters {
     } else {
       fmt.xLabelFormatter = (/** @type {any} */ val) => {
         if (Utils.isNumber(val)) {
+          const inferredNumericX =
+            w.axisFlags.isXNumeric &&
+            w.axisFlags.dataFormatXNumeric &&
+            w.config.xaxis.type !== 'datetime' &&
+            !w.globals.isBarHorizontal
           if (
             !w.config.xaxis.convertedCatToNumeric &&
-            w.config.xaxis.type === 'numeric'
+            (w.config.xaxis.type === 'numeric' || inferredNumericX)
           ) {
             if (Utils.isNumber(w.config.xaxis.decimalsInFloat)) {
               return val.toFixed(w.config.xaxis.decimalsInFloat)
--- a/src/modules/Formatters.js
+++ b/src/modules/Formatters.js
@@ -84,6 +84,12 @@ class Formatters {
   setLabelFormatters() {
     const w = this.w
     const fmt = w.formatters
+    const inferredNumericX =
+      w.axisFlags.isXNumeric &&
+      w.axisFlags.dataFormatXNumeric &&
+      w.config.xaxis.type !== 'datetime' &&
+      !w.config.xaxis.convertedCatToNumeric &&
+      !w.globals.isBarHorizontal
 
     fmt.xaxisTooltipFormatter = (/** @type {any} */ val) => {
       return this.defaultGeneralFormatter(val)
@@ -107,15 +113,18 @@ class Formatters {
     } else {
       fmt.xLabelFormatter = (/** @type {any} */ val) => {
         if (Utils.isNumber(val)) {
-          const inferredNumericX =
-            w.axisFlags.isXNumeric &&
-            w.axisFlags.dataFormatXNumeric &&
-            w.config.xaxis.type !== 'datetime' &&
-            !w.globals.isBarHorizontal
           if (
             !w.config.xaxis.convertedCatToNumeric &&
             (w.config.xaxis.type === 'numeric' || inferredNumericX)
           ) {
+            const ticks = w.globals.xAxisScale?.result
+            if (
+              inferredNumericX &&
+              Array.isArray(ticks) &&
+              ticks.every(Number.isInteger)
+            ) {
+              return val.toFixed(0)
+            }
             if (Utils.isNumber(w.config.xaxis.decimalsInFloat)) {
               return val.toFixed(w.config.xaxis.decimalsInFloat)
             } else {
@@ -150,6 +159,12 @@ class Formatters {
 
     if (typeof w.config.tooltip.x.formatter === 'function') {
       fmt.ttKeyFormatter = w.config.tooltip.x.formatter
+    } else if (
+      w.config.xaxis.labels.formatter === undefined &&
+      inferredNumericX
+    ) {
+      fmt.ttKeyFormatter = (/** @type {any} */ val) =>
+        Number.isInteger(val) ? val.toFixed(0) : fmt.xLabelFormatter(val)
     } else {
       fmt.ttKeyFormatter = fmt.xLabelFormatter
     }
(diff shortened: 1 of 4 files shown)


junedchhipa:
Right diagnosis, and it does fix the reported case. But routing inferred-numeric data through the default numeric formatter regresses label precision for a much wider population than the issue covers. A vertical bar chart with `x` values `2020, 2021, 2022` and no `xaxis.type` renders `2020.0 2021.0 2022.0` on this branch where `main` renders `2020 2021 2022`. Same for `2015..2024`, and a rangeBar with `x: 1..5` goes from `1 2 3 4 5` to `1.0 2.0 3.0 4.0 5.0`. Tooltip titles move with them, since `ttKeyFormatter` falls back to `xLabelFormatter`. The cause is the existing `diff / 10` decimals heuristic, which explicit `xaxis.type: 'numeric'` already suffers from, but this PR gives it a much bigger audience.

A guard ahead of that block covers it without giving up either of your improvements:

```js
const ticks = w.globals.xAxisScale?.result
if (Array.isArray(ticks) && ticks.every(Number.isInteger)) {
  return val.toFixed(0)
}
```

With that applied, `1, 2, 10` still labels `1.0 / 5.5 / 10.0` and `0.1, 0.2, 0.3` still labels `0.10 / 0.20 / 0.30`, all four of your new tests stay green, and the full unit suite passes.

Two smaller things.

`ticks = gl.dataPoints - 1` ignores the available width. A 250px chart with twelve seven-digit `x` values now draws all twelve labels with no rotation, and a 1200px chart with 29 points draws 29 where it drew 9. That rule is copied faithfully from the explicit-numeric path, so it is not new, but it now applies to charts that never opted into it.

An explicit `xaxis: { type: 'category' }` with numeric `{x, y}` data is reclassified as numeric as well, because `'category'` is the Options default and the two are indistinguishable at this layer. That may well be acceptable, since positioning was already by value there, but the PR body says string categories keep their existing path and this is the case where they do not.

The e2e suite is 313/313 on both `main` and this branch locally, so no shipped sample exercises any of the above. The year-labels case in particular needs a test of its own before this lands.

lovasoa:
Thanks for taking the time to look into my PR ! I'll add tests for all the cases you mentioned, then make the fix you suggest !

lovasoa:
@junedchhipa I added the regression coverage and pushed the follow-up fix. The new suite covers integer labels and tooltip keys, rangeBar, heatmap, 2-D data, explicit category axes, decimalsInFloat, scatter jitter, tickPlacement, and width-sensitive tick density. All 13 regression tests now pass, the focused x-axis/formatter run is 104/104, and the full unit suite is 3,364 passed with 6 expected failures. Could you please take another look?

PR #5305, opened 2026-09-11
fix(yaxis): skip text alignment for hidden y-axes

tarzan77cz:
## Description

`setYAxisTextAlignments()` throws `TypeError: Cannot read properties of null (reading 'getBoundingClientRect')` whenever a **hidden** y-axis has `labels.align` set.

`drawYaxis()` always creates the `.apexcharts-yaxis` group (so the `rel` index stays aligned with `w.config.yaxis`), but returns before adding `.apexcharts-yaxis-texts-g` when the axis is hidden:

https://github.com/apexcharts/apexcharts.js/blob/main/src/modules/axes/YAxis.js#L52-L56

```js
const elYaxis = graphics.group({ class: 'apexcharts-yaxis', rel: realIndex, ... })

if (this.axesUtils.isYAxisHidden(realIndex)) return elYaxis   // no texts group

const elYaxisTexts = graphics.group({ class: 'apexcharts-yaxis-texts-g' })
```

`setYAxisTextAlignments()` then iterates the `.apexcharts-yaxis` groups — hidden ones included — and dereferences the texts group unconditionally:

```js
const yAxisInner = w.dom.baseEl.querySelector(
  `.apexcharts-yaxis[rel='${index}'] .apexcharts-yaxis-texts-g`,
)
// ...
const rect = /** @type {Element} */ (yAxisInner).getBoundingClientRect()   // yAxisInner === null
```

The `/** @type {Element} */` casts are what currently silence the `@ts-check` warning about the possible `null`, so the type checker had already flagged this.

`isYAxisHidden()` returns `true` for `show: false`, for an axis whose series are all collapsed, and (with `showForNullSeries: false`) for all-null series — so this is reachable both from static config and at runtime, e.g. by clicking the legend until every series of one axis is collapsed on a chart that sets `labels.align`.

## Steps to Reproduce

```js
new ApexCharts(document.querySelector('#chart'), {
  chart: { type: 'line', height: 200 },
  series: [
    { name: 'a', data: [1, 2, 3] },
    { name: 'b', data: [3, 2, 1] },
  ],
  yaxis: [
    { labels: { align: 'left' } },
    { show: false, labels: { align: 'left' } },
  ],
}).render()
```

## Expected Behavior

The chart renders and the visible axis gets its label alignment; hidden axes are skipped.

## Actual Behavior

`setYAxisTextAlignments()` throws on the hidden axis, which aborts the remaining alignment work in the same `forEach` and surfaces as an uncaught error in the console on every render.

## Fix

Skip the entry when the texts group is absent. This mirrors the guard already used a few methods above in the same file, in `yAxisTitleRotate()`:

```js
const yAxisLabelsCoord = elYAxisLabelsWrap
  ? elYAxisLabelsWrap.getBoundingClientRect()
  : { width: 0, height: 0 }
```

With the early return, `yAxisInner` narrows to `Element`, so the four `/** @type {Element} */` casts are no longer needed and are removed.

## Notes

Found while debugging a two-series chart in [apexcharts-card](https://github.com/RomRider/apexcharts-card) (Home Assistant), where the second y-axis ends up hidden while inheriting `labels.align: 'left'` from the first — three uncaught exceptions per page load, with the chart itself rendering correctly.

### DIFF
--- a/src/modules/axes/YAxis.js
+++ b/src/modules/axes/YAxis.js
@@ -561,32 +561,31 @@ export default class YAxis {
         const yAxisInner = w.dom.baseEl.querySelector(
           `.apexcharts-yaxis[rel='${index}'] .apexcharts-yaxis-texts-g`,
         )
+        // A hidden axis still renders its .apexcharts-yaxis group, so it is part
+        // of the collection above and keeps the rel index aligned, but drawYaxis
+        // returns before adding the texts group — there is nothing to align.
+        if (!yAxisInner) return
+
         const yAxisTexts = Array.from(
           w.dom.baseEl.querySelectorAll(
             `.apexcharts-yaxis[rel='${index}'] .apexcharts-yaxis-label`,
           ),
         )
-        const rect = /** @type {Element} */ (yAxisInner).getBoundingClientRect()
+        const rect = yAxisInner.getBoundingClientRect()
 
         yAxisTexts.forEach((label) => {
           label.setAttribute('text-anchor', yaxe.labels.align)
         })
 
         if (yaxe.labels.align === 'left' && !yaxe.opposite) {
-          ;/** @type {Element} */ (yAxisInner).setAttribute(
-            'transform',
-            `translate(-${rect.width}, 0)`,
-          )
+          yAxisInner.setAttribute('transform', `translate(-${rect.width}, 0)`)
         } else if (yaxe.labels.align === 'center') {
-          ;/** @type {Element} */ (yAxisInner).setAttribute(
+          yAxisInner.setAttribute(
             'transform',
             `translate(${(rect.width / 2) * (!yaxe.opposite ? -1 : 1)}, 0)`,
           )
         } else if (yaxe.labels.align === 'right' && yaxe.opposite) {
-          ;/** @type {Element} */ (yAxisInner).setAttribute(
-            'transform',
-            `translate(${rect.width}, 0)`,
-          )
+          yAxisInner.setAttribute('transform', `translate(${rect.width}, 0)`)
         }
       }
     })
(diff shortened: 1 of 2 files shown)


junedchhipa:
Verified the bug and the fix. Worth noting it is a bit worse than the description: `render()` *rejects*, so everything in `mount()` after the alignment call is skipped, and on a three-axis chart with the hidden axis at `rel=0` the two visible axes end up with no alignment transform at all. Your guard fixes both.

One thing I would fold in while you are already in those lines. The `forEach` still takes its index from the DOM collection, but `Axes.drawAxis` skips `globals.ignoreYAxisIndexes` entirely, so when every series of an axis is collapsed (one of the runtime cases you cite) the collection is *shorter* than `config.yaxis` and the last axis is never visited. With three axes and the first axis's series collapsed, `rel=2` keeps its alignment on neither `main` nor this branch. Iterating `w.config.yaxis` fixes that too, and lets the `getElementsByClassName` lookup go away:

```js
setYAxisTextAlignments() {
  const w = this.w

  w.config.yaxis.forEach((yaxe, index) => {
    if (yaxe && !yaxe.floating && yaxe.labels.align !== undefined) {
      // ... unchanged, including your new guard
```

Full unit suite and all 313 e2e samples pass with that in place.

Could you also add a regression test? Asserting that `render()` resolves for the two-axis config in your repro would have caught this. And a heads-up that CI has not run on this branch yet, since it is your first PR here, so I will approve the workflow. Locally: 3367 unit tests pass, e2e 313/313, lint clean, typecheck unchanged at the pre-existing 23.

tarzan77cz:
Folded in iterating `w.config.yaxis` instead of the DOM collection, kept the missing texts-group guard, and added a regression that `render()` resolves for the two-axis repro. Also covers the collapsed-first-axis case so the last axis still gets its alignment transform.

Issue #4469, opened 2024-05-14
Feature Request: Enable Interactive Tooltips in ApexCharts


FADL285:
## Summary

Currently, tooltips in ApexCharts hide when the mouse moves away from the chart marker, making it difficult to interact with elements within the tooltip. I propose adding an option to make tooltips interactive, allowing them to remain visible when hovered over. This would enable users to select or interact with custom content inside the tooltip without it disappearing.

## API Changes

Add a new configuration option in the `tooltip` settings:

```javascript
tooltip: {
  interactive: true // Default value: false
}
```

### Example Usage

```javascript
var options = {
  chart: {
    type: 'scatter'
  },
  series: [{
    name: 'Sample Data',
    data: [[1, 2], [2, 3], [3, 4], [4, 5], [5, 6]]
  }],
  tooltip: {
    enabled: true,
    interactive: true,
    custom: function({ seriesIndex, dataPointIndex, w }) {
      const seriesItem = w.config.series[seriesIndex];
      return `
        <div>
          <p>${seriesItem.name}</p>
          <a href="${seriesItem.metadata.url}" target="_blank">More Info</a>
        </div>
      `;
    }
  }
};
```

## Intended Use Case

This feature would be particularly useful for dashboards and data-driven applications where users need to interact with additional information or actions provided within the tooltip. For example, in financial charts, users could click on links for more detailed reports, or in e-commerce analytics, they could view and interact with product details directly from the tooltip. This enhancement will improve user experience by making the tooltip content more accessible and functional.

PR #5307, opened 2026-09-12
feat: add interactive tooltips

lovasoa:
Hi ! I am working on [SQLPage](https://sql-page.com), which bundles apexcharts. I recently added the ability to add links to datapoints in charts, but current apexcharts auto-closes tooltips when you try to hover them, which makes clicking them difficult.

This pr adds a `tooltip.interactive` option, defaulting to `false`.
When set to true, it keeps the tooltip open long enough for the pointer to cross the point-to-tooltip gap.

Usage : 
```js
tooltip: {
  interactive: true,
  custom: () => '<a href="/details">View details</a>'
}
```

Implements #4469.

https://github.com/user-attachments/assets/b9daa981-918a-4a9f-9bec-e11e4cdd30bc

### DIFF
--- a/src/assets/apexcharts.css
+++ b/src/assets/apexcharts.css
@@ -430,6 +430,9 @@ rect.legend-mouseover-inactive,
   opacity: 1;
   transition: opacity .12s ease
 }
+.apexcharts-tooltip.apexcharts-tooltip-interactive {
+  pointer-events: auto;
+}
 .apexcharts-tooltip.apexcharts-active[data-positioned="true"] {
   transition: opacity .12s ease, left .16s ease-out, top .16s ease-out
 }
--- a/src/modules/settings/Options.js
+++ b/src/modules/settings/Options.js
@@ -2480,6 +2480,7 @@ export default class Options {
         followCursor: false, // when disabled, the tooltip will show on top of the series instead of mouse position
         intersect: false, // when enabled, tooltip will only show when user directly hovers over point
         inverseOrder: false,
+        interactive: false,
         arrow: true,
         // One tight line instead of a card: the x label sits inline before
         // the value, the marker goes, the padding and font shrink. For panels
--- a/src/modules/tooltip/Tooltip.js
+++ b/src/modules/tooltip/Tooltip.js
@@ -11,6 +11,8 @@ import XAxis from './../axes/XAxis'
 import Utils from './Utils'
 import { isCustom } from '../ChartFactory'
 
+const INTERACTIVE_TOOLTIP_HIDE_DELAY = 150
+
 /**
  * ApexCharts Core Tooltip Class to handle the tooltip generation.
  *
@@ -90,6 +92,8 @@ export default class Tooltip {
     this.seriesBound = null
     /** @type {ReturnType<typeof setTimeout> | undefined} */
     this.seriesHoverTimeout = undefined
+    /** @type {ReturnType<typeof setTimeout> | undefined} */
+    this.interactiveHideTimeout = undefined
     /** @type {number} */
     this.clientX = 0
     /** @type {number} */
@@ -231,6 +235,22 @@ export default class Tooltip {
       'div',
     )
     tooltipEl.classList.add('apexcharts-tooltip')
+    if (this.tConfig.interactive) {
+      tooltipEl.classList.add('apexcharts-tooltip-interactive')
+      tooltipEl.addEventListener(
+        'mouseenter',
+        () => clearTimeout(this.interactiveHideTimeout),
+        { passive: true },
+      )
+      tooltipEl.addEventListener(
+        'mouseleave',
+        () => {
+          clearTimeout(this.interactiveHideTimeout)
+          this.handleMouseOut({ tooltipEl })
+        },
+        { passive: true },
+      )
+    }
     if (w.config.tooltip.cssClass) {
       tooltipEl.classList.add(w.config.tooltip.cssClass)
     }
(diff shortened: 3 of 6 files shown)


junedchhipa:
Thanks for this. The opt-in shape is right, and the interaction test is welcome.

One thing to settle before I merge: `followCursor: true` moves the tooltip with the pointer, so an interactive tooltip can never actually be reached. Today the two can be set together and the result is just a tooltip that runs away from you. Could you either ignore `followCursor` while `interactive` is on, or note in the option comment that the two do not combine? Either way is fine, whichever reads better to you.

Worth mentioning in that comment too that this is aimed at `custom` tooltips with clickable content, since that is where it earns its keep.

lovasoa:
Implemented in 0547a8dbb. tooltip.interactive now takes precedence over followCursor across tooltip positioning paths, while preserving the configured value for when interactivity is later disabled. I also documented that this supports clickable custom-tooltip content and added a browser test covering both options together.

junedchhipa (changes requested):
Thanks for the follow-up. The `isFollowCursor()` precedence is the right shape, the types and consumer test are updated, `typecheck:types` passes and `tsc` reports the same 23 pre-existing errors as `main`. The three browser tests pass here against a freshly built dist.

Two blockers remain.

### 1. The red CI is this branch: 14 unit tests fail

`tests/unit/tooltip.spec.js` and `tests/unit/tooltip-positioning.spec.js` drive `Position` with a hand-built `ttCtx` double that carries `w` but no methods, so every call site now throws `TypeError: ttCtx.isFollowCursor is not a function`. You can add the method to the doubles, but I would rather the check not need a method on the context at all: a helper that takes `w` (for example in `tooltip/Utils.js`, returning `w.config.tooltip.followCursor && !w.config.tooltip.interactive`) keeps `Position` and `Intersect` working for any caller that has `w`, which is what those tests already are.

### 2. Grouped charts leave the sibling tooltip stuck

The early `return` on `mouseout` in `onSeriesHover` skips `seriesHover`, and `seriesHover` is what fans the event out to the other charts in a `chart.group`. I probed two grouped line charts with `interactive: true`: hover one, move the pointer off both, wait 500ms. The hovered chart's tooltip hides, the sibling's keeps `apexcharts-active`. With `interactive: false` both hide.

The simplest fix is to let the deferred timer run the normal path, `this.seriesHover(opt, e)`, instead of calling `handleMouseOut(opt)` directly, so the siblings receive the same `mouseout` they do today. A `chart.group` case in `tooltip-interactive.spec.js` would lock it in.

### Minor

The `interactive: false` line in `Options.js` has no comment while the note about `followCursor` and clickable `custom` content lives only in the `.d.ts`. Please mirror it there, since `Options.js` is where these are documented for the source.

lovasoa:
Implemented the requested fixes in ebda40f6b: the follow-cursor check now uses w, grouped interactive mouseout propagates correctly, and the source option comment is updated. All CI checks are now passing; could you please take another look and review when convenient?

junedchhipa (changes requested):
The source changes are right, and I have confirmed both of them. `TooltipUtils.isFollowCursor(w)` is the shape I was after, and the unit specs are back to 118/118 with the full suite at 3377 passed plus the 6 expected fails. The grouped fix works: I drove two grouped charts in Chromium, hovered one and moved the pointer off both, and with `interactive: true` both tooltips now hide where the sibling used to stay stuck.

Two things before this lands.

### 1. The new grouped test does not exercise the grouped path, and fails

`hides synchronized grouped tooltips after mouseout` is red for me on `ebda40f6b`, three runs out of three. Note that `test:ci` is `e2e:ci` plus `unit`, so the Playwright suite does not run in CI. A green CI does not cover this file. Please run `yarn interaction` locally.

Two independent causes:

- **`.hover()` never reaches the chart.** Playwright aims at the centre of the bounding box of `#left .apexcharts-series path`, and for a thin diagonal line that point is empty space, so the hit lands on the full-size `<foreignObject>` that `Core.js` puts behind the plot. Use a coordinate move over `.apexcharts-grid`, which is what the other specs and the suite's own hover helper do.
- **The charts never form a group.** Instances register in `Apex._chartInstances` only `if (this.w.config.chart.id)`, and `mountGrouped` sets `group` but no `id`. So `getSyncedCharts()` returns just the one chart. With the hover fixed, the test then fails on the first assertion instead: `#right` never becomes active. Give each chart its own `chart.id` and it exercises the real path.

With both corrected the scenario passes, so the test is one line of setup away from guarding your fix.

### 2. The same leak survives on the tooltip's own exit

Deferring to `seriesHover` fixes the case where the pointer leaves the chart. It does not cover the case this feature exists for: move into the tooltip, then leave it. The `mouseleave` listener on the tooltip element calls `this.handleMouseOut({ tooltipEl })` directly, which is the same bypass one entry point over. Measured on two grouped charts with `interactive: true`, leaving the tooltip outward: the hovered chart hides, the sibling keeps `apexcharts-active`.

Mirroring the close to the group fixes it. I built this and re-ran the probe, and both tooltips hide:

```js
tooltipEl.addEventListener('mouseleave', () => {
  clearTimeout(this.interactiveHideTimeout)
  this.handleMouseOut({ tooltipEl })
  if (w.config.chart.group) {
    this.ctx.getGroupedCharts().forEach((ch) => {
      const el = this.getElTooltip(ch)
      if (el) ch.w.globals.tooltip.handleMouseOut({ tooltipEl: el })
    })
  }
})
```

Worth a second case in the spec covering that exit, since it is the one users will take.

### Looks good

The `Options.js` comment reads well. No regressions: the other 41 tooltip, shadow-dom and syncing interaction tests pass, and the branch merges into `main` cleanly despite being 49 commits behind.

lovasoa:
@junedchhipa : can we talk human to human ? 

I didn't realize end to end tests didn't run on ci, do you want a pr that fixes that ?

junedchhipa:
Yes, fair. Those reviews got long and turned into a checklist. Sorry about that.

And yes please on the CI PR, that would be very welcome. The interaction suite is 584 tests and green on main today, so it should come down to a Playwright install step and a job, with no backlog to clear first.

junedchhipa (approved):
Both sorted, thanks. Leaving the tooltip now closes the siblings too, and the grouped spec actually forms a group so it would catch a regression.

Good to merge.

Issue #3481, opened 2022-11-10
CSV Export does not work properly when one of the serie misses a data point


valleen:
### Steps to Reproduce

1. go to https://codepen.io/nycoss/pen/ExRWboX
2. export to csv
3. only data common to both series are exported and relation date/value is broken

### Expected Behavior
I want to see all dates and a blank or zero for the specific serie that misses one data point
### Actual Behavior
data vector is incorrect

juan-carvajal:
Same problem here

Lennert0d:
I have also experienced this issue.

PR #5308, opened 2026-09-12
fix(exports): key unequal-x CSV rows without `Object.prototype`

81reap:
`handleUnequalXValues` grouped rows in a plain unvalidated object keyed by the category. So a category named `toString` threw `values is not iterable`; while one named `__proto__` wrote through to `Object.prototype`, leaving every object on the page with an `0` property.

Now it is keyed in a Map by the category name.

# New Pull Request

I tried searching and there doesn't seem to be an open issue for this. I made a codepen to show a before and after. https://codepen.io/editor/81reap/pen/01a097ef-876f-7276-9108-4369f365387f

#3481 sounds similar but I'm not entirely sure if its the same issue

## Type of change

## Checklist:

## AI Usage 

Hi 👋🏽 I'm a human. I don't let agents post or act fully autonomously on my behalf. But I did use claude code to help make the fix and codepen. I have reviewed the code and it makes sense to me but I'm happy to make edits. :)

### DIFF
--- a/src/modules/Exports.js
+++ b/src/modules/Exports.js
@@ -999,7 +999,9 @@ class Exports {
 
     const handleUnequalXValues = () => {
       const categories = new Set()
-      const data = {}
+      // Map stops a plain object category named `__proto__` from writing
+      /** @type {Map<string, string[]>} */
+      const byCategory = new Map()
 
       /**
        * @param {Record<string, any>} s
@@ -1020,13 +1022,13 @@ class Exports {
           } else {
             return
           }
-          if (!(/** @type {Record<string,any>} */ (data)[cat])) {
-            ;/** @type {Record<string,any>} */ (data)[cat] = Array(
-              series.length,
-            ).fill('')
+          const key = String(cat)
+          let values = byCategory.get(key)
+          if (!values) {
+            values = Array(series.length).fill('')
+            byCategory.set(key, values)
           }
-          ;/** @type {Record<string,any>} */ (data)[cat][sI] =
-            getFormattedValue(value)
+          values[sI] = getFormattedValue(value)
           categories.add(cat)
         })
       })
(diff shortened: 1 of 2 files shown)


junedchhipa:
Thanks, this is a good catch and the fix is the right shape. Both symptoms reproduce on `main` exactly as you describe:

```
category named "toString"   → throws "values is not iterable"
category named "__proto__"  → throws, and sets Object.prototype[0] = 1
```

The second one is the serious half. The pollution outlives the failed export, so `({})[0] === 1` for every object on the page afterwards. While writing my own probe I watched it break unrelated later tests in the same file with `fn is not a function`, which is a fair preview of what it would do to a host app.

I checked the fix is complete rather than just sufficient for the two names in your tests. `byCategory` is the only data-keyed plain object left in `exportToCSV`, and I ran ten prototype-named categories (`toString`, `__proto__`, `constructor`, `valueOf`, `hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable`, `toLocaleString`, `__defineGetter__`, `length`) through both the `{x, y}` and the 2D-array formats: all ten export correctly and `Object.prototype` stays clean. Both of your new tests fail on `main` and pass here.

One thing I looked at and decided is not yours to fix: the Map keys on `String(cat)` while `categories` stays a Set of raw values, so a chart carrying both numeric `1` and string `'1'` emits a duplicate row. I isolated that on both branches and the output is identical, because the plain object coerced its keys the same way. Pre-existing and unchanged, so it is out of scope here.

Verified locally: unit suite passes, e2e 313/313 samples, lint clean, typecheck at the same pre-existing count, and a full `yarn build` succeeds. Noting that last one because CI had not run on this branch yet, first PR from a new contributor.

And thanks for the note about tooling. Merging.

PR #5310, opened 2026-09-14
fix(series): keep points appended to a hidden series

### LINKED ISSUE #3200: appendData doesn't work for hidden series
### Description

With Multi charts, appendData to collapsed series would be lost.
I have a line chart with multiple lines and updating d every 30s, so I would like to call appendData for all the series.
`chart.appendData([ { name: 'series_1',  data: [1] }, { name: 'series_2', data: [2] } ])` if series_2 is toggled hidden, the data point will not be added to that series, when toggle visible again, all the data appended during time series_2 is not visible will be lost.

### Steps to Reproduce

1.  create multi line chart
2.  toggle some series hidden 
3.  call appendData
4.  toggle the hidden series to be visible
5. The appended Data only added to the ones that were visible at that point. 

### Expected Behavior

appendData should append to series wether it is hidden or visible

### Actual Behavior
appendData doesn't work for hidden series
### Reproduction Link

https://codepen.io/just_yuki/pen/qBxJMJm

lbesecker195:
# New Pull Request

`appendData()` loses the points it appends to a series that is hidden from the legend.

A hidden axis-chart series is kept in two places. `config.series[i].data` is emptied, and the original data is stored on the collapse record in `globals.collapsedSeries` (or `ancillaryCollapsedSeries`), which `showSeries()` restores from. `appendData()` only pushed into `config.series`, so for the hidden series:

- the points went into the emptied row and were dropped when the series was shown again, because the rise restores the record's data;
- while hidden, that row held just the new points, which were parsed into the axis range.

With series `A: [1, 2, 3]`, `B: [4, 5, 6]`, then `hideSeries('B')` and `appendData([{ data: [7] }, { data: [8] }])`, main gives:

| | main | this PR |
|---|---|---|
| `config.series[1].data` while hidden | `[8]` | `[]` |
| `globals.maxY` while hidden | 8 | 7 |
| B after `showSeries('B')` | `[4, 5, 6]` | `[4, 5, 6, 8]` |

The fix appends a hidden series' points to its collapse record instead of the config row. The same `trimStreamingSeries()` bound is applied to those records, so a stream with `maxPoints` or `xaxis.range` doesn't grow a hidden series without limit. Visible series, non-axis charts, and the histogram, waterfall, dumbbell and streamgraph path at the top of `appendData()` are unchanged.

Fixes #3200

## Type of change

## Tests

Two tests in `tests/unit/series-collapse-across-update.spec.js`, next to the tests from 2d67958 for the same records:

- `appendData` while B is hidden: A gets the point, B's row stays empty, `maxY` is 7, and `showSeries('B')` restores `[4, 5, 6, 8]`.
- With `chart.streaming: { enabled: true, maxPoints: 4 }`: after appending two points to each, both series hold the last 4 once B is shown.

Without the change to `src/apexcharts.js`, both fail, with `[8]` for the hidden row and `[4, 5, 6]` after the rise.

## Checklist:

Details:

- `npx vitest run tests/unit/`: 144 files, 3415 passed and 6 expected fail. Main has 3413 passed and the same 6 expected fail.
- `npx eslint` on the changed files passes.
- `npm run typecheck` reports the same 24 errors on main and on this branch, none in the changed lines.
- I did not run the `e2e` screenshot suite.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


junedchhipa (changes requested):
Thanks, the diagnosis is right and the table in the description matches what I measure on `main`. Two things before this can land.

### 1. Rebase: the loop body changed on `main`

Three commits touched `appendData` after this branched (18868c003, 79eee1239, 1ee0abb71), and the conflict is exactly the hunk you changed. `main` now replaces the array (`dstSerie.data = dstSerie.data.concat(srcSerie.data)`) instead of pushing into it, because the snapshots share these arrays and growing them in place is the one mutation they cannot be shielded from. The record append needs the same shape:

```js
const record = collapsedRecords.find((c) => c.index === i)
if (record && Array.isArray(record.data)) {
  record.data = record.data.concat(srcSerie.data)
} else {
  dstSerie.data = dstSerie.data.concat(srcSerie.data)
}
```

With that resolution your two tests pass along with the rest of the file (8/8), and the streaming, append, series, legend and reset suites stay green (122 tests).

### 2. The baseline still loses the hidden series

`appendData` passes `overwriteInitialSeries: true`, and the parse then snapshots `initialSeries` from `config.series`, where the hidden row is empty. So after `hideSeries('B')` and the append, the baseline for `B` is `[]`. Two readers get the wrong answer from that: `resetSeries()` now restores `B` as empty for the life of the chart (`main` gives `[8]` there, so both are wrong, this one loses more), and `isInitialSeriesSameLen()` reads false after `showSeries('B')`, which drops the shared tooltip to the nearest single series.

The record you are appending to is the right source for that row. #5313 adds a repair in `create()` that puts a collapsed row's data back into the baseline for series declared hidden in the config; generalising it to every collapse record covers this case too, and I have suggested the record as the source over there so the two compose. Either way, please add a test that asserts `initialSeries[1].data` and `resetSeries()` after the append, since the current tests only read `config.series`.

Nit: `Fixes #3200` is in the body, good, that is what closes the issue.

PR #5311, opened 2026-09-14
fix(bar): data labels on a 100% stacked combo

### LINKED ISSUE #2429: Received 'TypeError: Cannot read property 'toString' of undefined' when try to mix bar and line at stacked chart with stackedType="100%"
# Bug report

Got "TypeError: Cannot read property 'toString' of undefined" when trying to plot stacked chart mixed line and bar with 
stackType set to '100%'

## Codepen

https://codepen.io/mvtm/pen/VwpbpYw

## Explanation

I expect to see stacked bar and line over it
- What is happening instead?
Partially drawn plot without axes, lines etc
``````
apexcharts:6 Uncaught (in promise) TypeError: Cannot read property 'toString' of undefined
    at Function.value (apexcharts:6)
    at c (apexcharts:6)
    at s.value (apexcharts:6)
    at t.value (apexcharts:6)
    at t.value (apexcharts:14)
    at t.create (apexcharts:6)
    at apexcharts:14
    at new Promise (<anonymous>)
    at t.value (apexcharts:14)
    at pen.js:51
```

gioboa:
Closes #2429

# New Pull Request

Fixes data-label crash on 100% stacked combos: bars read a share of the bar total, line keeps raw values. Closes #2429.

## Type of change

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand ar
- [x] I have added tests that prove my fix is effective or that my fe


junedchhipa (changes requested):
Thanks, the decision that only the bar series are slices of the 100% total is the right one, and the write-up in the comments will save the next person a lot of time.

What I checked: the new unit spec is 5/6 red on `main` (`formatter is not a function`) and 6/6 green here; the bar, stacked and data-label suites stay green; the browser spec passes 3/3 against a freshly built dist; CI and the e2e snapshots are green.

Two small changes before merge.

### 1. Drop the `chart.type === 'bar'` branch

A combo can also be typed `bar` with a `line` series mixed in. That path still gets the bare percent formatter for every series, so the line's raw values render as `180%` / `290%`. Without the branch, the combo-aware formatter handles it: bars read `67%` / `33%`, the line reads `180` / `290`. Pure bar charts do not change, because `columnSeries.i` covers every series there. The new spec and the bar suites stay green with the branch removed (41 tests).

### 2. Avoid `new Options().init()` for one function

That builds the whole default config to fetch the plain formatter. The default is just `val !== null ? val : ''`, so a local function does the same job without the allocation.

One thing to state as a deliberate choice rather than a side effect: a pure 100% stacked area now labels each point with its raw value over a 0 to 100 axis. That is a big improvement on a crash and the spec pins it, so fine by me, just worth a line in the PR description.

gioboa (commented):
Thanks for your review @junedchhipa 
[7831fbd](https://github.com/apexcharts/apexcharts.js/pull/5311/commits/7831fbd8e7fc9e923752ec49207a7711527583fa) is fixing it

junedchhipa (approved):
Both addressed, thanks, and the test you added for the bar-typed combo is the right one to have.

Verified on `7831fbd8e`: that test is red on `1312f99ad` and green here, so it guards the case rather than just describing it. Full unit suite 146 files, 3429 passed and the 6 expected fails. eslint clean, merges into `main` with no conflict.

The Playwright suite does not run in CI, so I ran it here too: the spec you modified passes 3/3, and 96 related stacked, bar and data-label interaction tests pass.

Good to merge.

PR #5313, opened 2026-09-20
fix(series): keep the baseline of a series declared hidden in the config

mrash:
`ApexCharts.create()` collapses every series carrying `hidden: true` before `parseData()` runs, so the baseline that parse snapshots into `globals.initialSeries` holds `data: []` for each of them. Until 7.4.0 the first legend interaction repaired that by accident, because `parseData()` re-snapshotted unconditionally. It no longer does: the legend's own updates pass `overwriteInitialSeries: false` so that an internal re-render keeps the baseline it was given (#5283), and the emptied rows are now permanent.

Two readers get the wrong answer:

- `Series.resetSeries()` clones the baseline into `config.series`, so a series declared hidden comes back from a reset EMPTY and its data is gone for the life of the chart;
- `Tooltip.tooltipUtil.isInitialSeriesSameLen()` filters out collapsed rows but measures the rest, so the moment the viewer un-hides one from the legend its length of 0 is compared against its siblings'. The check fails, `handleStickyCapturedSeries()` falls to `create(..., false)`, and every shared tooltip on the chart silently drops to the single series nearest the cursor. Re-hiding the series filters it out again and the tooltip comes back, which is what makes it read as a tooltip bug rather than a data one.

Restore the caller's own data for exactly the rows the `hidden` flag emptied. The raw-stash baselines (histogram, dumbbell, streamgraph, waterfall, treemap and the dataReducer window) are left as `parseData()` wrote them, and a row whose input is itself empty is skipped, so the repair can never turn a good baseline into an empty one.

Sibling of #5118, which is the same hazard one snapshot over: `initialConfig` lost a collapsed series' data for a different reason and was fixed by copying the series objects at capture time.

6 tests in tests/unit/config-hidden-series-baseline.spec.js; 4 of them fail on main. Full unit suite 151 files / 3495 tests green, eslint clean, and `tsc --noEmit` reports the same 29 pre-existing errors as main.

## Checklist:


junedchhipa (approved):
Thanks, this is a clean diagnosis and the fix is scoped exactly right. Approving with one suggestion.

What I checked: the new spec is 4/6 red on `main` and 6/6 green here; the tooltip, series, legend, history and initial-config suites stay green (251 tests); the `resetSeries` assertion still holds after the update settles, so it is not a timing artefact. CI is green and the branch is one commit behind `main`.

### Suggestion: source the repair from the collapse record

`getSeriesAfterCollapsing` already stores the hidden row's data on its record in `gl.collapsedSeries` / `gl.ancillaryCollapsedSeries`. At init that record and `ser[i].data` are the same thing, but they diverge on the one re-entry that also passes `overwriteInitialSeries: true`: `appendData()`. The `hidden` flag is still on the config row, so `hiddenAtInit` fires again, and `ser[i].data` is then only the points just appended.

With `B` declared hidden and `appendData([{ data: [7] }, { data: [8] }])`, the baseline for `B` becomes `[8]` on this branch, while the record still holds `[4, 5, 6]`. Reading from the record instead:

```js
const records = gl.collapsedSeries.concat(gl.ancillaryCollapsedSeries)
gl.initialSeries = baseline.map((s, i) => {
  if (hiddenAtInit.indexOf(i) === -1 || !Utils.isObject(s)) return s
  const rec = records.find((c) => c.index === i)
  return rec && Array.isArray(rec.data) && rec.data.length
    ? { ...s, data: rec.data }
    : s
})
```

keeps the spec 6/6 and gives `[4, 5, 6]` in that case. It also composes with #5310, which appends new points into that same record, so the two together leave the baseline right after an append as well.

Happy to merge once that is in, or as-is if you would rather keep the append case for a follow-up.

mrash:
Thanks - I've incorporated your suggestion above. Two things added: 

  - .slice() the record's array instead of aliasing it. The initialSeries setter keeps a shallow copy and _initialSeriesPeek hands it straight to the tooltip's same-length check — sharing the array would let a later in-place edit of the
    record reach a captured baseline, which is the one hazard that snapshot exists to rule out. .slice() is already the house idiom at this boundary (getSeriesAfterCollapsing, riseCollapsedSeries).
  - Array.isArray(rec.data) is load-bearing, not just defensive. A non-axis collapse records one slice's value, not a series' rows. That check stands in for the old ser[i].data shape test.
  
  Two tests added: the append case (red before — [8] where [4,5,6] is expected) and an aliasing guard. Spec is 8/8.
  Upstream: rebase, not merge
  
  Rebased onto upstream/main (88c7e932e), clean, no conflicts. Rebase is right here — the two upstream commits (Animations.js/Globals.js/internal.d.ts, release-notes) have zero file overlap with ours, and upstream lands PRs with Merge 
  pull request commits, so a reverse-merge inside the branch would just be noise in the PR diff.
  
  Verification on the rebased tree: unit suite 152 files / 3527 tests green, eslint clean, tsc --noEmit 23 errors — the same 23 as upstream/main (I measured main's count in a throwaway worktree; note the PR body's "29" was stale from
  the 7.5.0 session).

Also, regarding the force push:

  Here's the actual state:
  
                                ┌─ 2dc2ac8de  ← origin (GitHub fork, what PR #5313 shows)
  e56e81d64 (v7.5.1) ───────────┤
                                └─ 60671bbf5 ─ 88c7e932e ─ 82344e137 ─ e697d0044  ← yours
                                   (upstream/main moved)   (rebased)    (review fix)
  
  Merge base is e56e81d64. Ours is 4 ahead / 1 "behind" — and that 1 remote-only commit is just the pre-rebase version of your own work. I checked the patch-ids:
  
  remote 2dc2ac8de: f2d29307db91b0d215923cc38b20fb5a37afa63f
  local  82344e137: f2d29307db91b0d215923cc38b20fb5a37afa63f   ← identical
  
  So 2dc2ac8de is fully contained in your branch as 82344e137. Nothing on the remote would be lost. A force-push is correct, not destructive.

junedchhipa (approved):
Both deviations are right, and I prefer them to what I suggested.

The `.slice()` is the correct call for the reason you give. I checked that the guard you added for it can actually fail: with the `.slice()` removed and everything else unchanged, `does not alias the collapse record into the baseline` goes red. A guard that cannot fail would have been worse than none, so thanks for making that one real.

`Array.isArray(rec.data)` doing the shape test is a better answer than the `ser[i].data` truthiness chain it replaces, because it says what it is actually checking.

Verified on `e697d0044`:

- the spec is 8/8, and `keeps the whole series in the baseline across an appendData` is red on `82344e137`, so the new behaviour is guarded
- full unit suite 152 files, 3521 passed and the 6 expected fails, which matches your numbers exactly
- eslint clean, merges into `main` with no conflict

The rebase was the right call, and thanks for flagging the stale error count in the description rather than leaving me to find it.

Good to merge from my side.

PR #5314, opened 2026-09-22
fix(range): let rangeArea charts scale the y axis to their data

### LINKED ISSUE #4999: autoScaleYaxis does not work on rangeArea chart
### Description

When using rangeArea chart, y axis auto scaling doesn't work. The Y axis always starts from 0, unlike on other chart types.

### Steps to Reproduce

1.  Open: https://codepen.io/Esko-the-reactor/pen/mydGxmP
2.  Observe how the smallest value is 100k, but the Y axis starts from 0

### Expected Behavior
Y axis should start from ~100k in the above example and not from 0.
Here is same data on a line chart which scales correctly: https://codepen.io/Esko-the-reactor/pen/EaxeEGq

### Actual Behavior
Y axis is not scaled properly to the input data and starts from 0.

### Screenshots
This screenshot shows the error case with rangeArea chart:
Image

This screenshot shows how line chart is scaled properly with the same data:
 
Image

### Reproduction Link

 https://codepen.io/Esko-the-reactor/pen/mydGxmP

mmilanovic4:
Fixes #4999

A `rangeArea` chart always started its y axis at 0, so ranges far from zero
(e.g. 100k–115k) were squashed into a thin band and `chart.zoom.autoScaleYaxis`
had no visible effect.

`Range.setYRange()` only lets line/area/scatter/candlestick/boxPlot/violin and
vertical rangeBar charts take their lower bound from the data. `rangeArea` was
missing from that list, so it fell through to the bar-style branch that merges
in `Number.MIN_VALUE` and pins the axis to 0. This adds `rangeArea` to the list.

Tests: `tests/unit/issue-4999-rangearea-autoscale.spec.js` (3 of 4 cases fail
without the fix): axis no longer pinned to 0, matches a line chart over the same
lows, rescales to a zoomed window with autoScaleYaxis, and still includes ranges
that dip below zero. Full unit suite passes; e2e snapshots for the existing
rangeArea samples are pixel-identical before/after.

## Type of change


junedchhipa (approved):
Clean diagnosis, and the minimal fix for it. Thanks for the oracle test against a line chart over the same lows, that is the right way to pin this.

The mechanism holds up: `minY` only moves off `Number.MIN_VALUE` when some value is negative, so for all-positive data the `else` branch assigns the sentinel and the axis starts at 0, while the listed types take `lowestY` instead. Adding `rangeArea` to the list is the whole of it.

Verified on `f0daf94ae`:

- your spec is 4/4 here and 3/4 red against `main`'s `Range.js`, so it guards the behaviour
- full unit suite 154 files, 3532 passed plus the 6 expected fails
- e2e 316 samples all pass, which confirms your pixel-identical claim
- eslint clean, merges into `main` with no conflict

The case I most wanted to check was `streamgraph`, which is a `rangeArea` alias in `TYPE_ALIASES` and so reaches this branch too. It is unaffected, for a reason worth writing down: the wiggle baseline always straddles zero, so `minY` is already a real negative by the time `setYRange` runs and the new guard (`gl.minY === Number.MIN_VALUE`) never fires. I measured it both ways, all-positive input and values around 100k, and `minY` / `maxY` are byte-identical before and after. 79 streamgraph, range, zoom and brush interaction tests pass as well.

One note for whoever touches this next: adding a type to that list silently opts in every alias that routes through it. Benign here, but it is not obvious from the code.

CI has not run yet, since a first-time contributor's workflow run needs a maintainer to approve it.

Issue #3579, opened 2023-01-02
horizontal bar chart - total dataLabel overflow on stackType 100%


delphi-sucks:
### Description

When creating a horizontal 100% stacked bar chart with a total-dataLabel, the dataLabel gets cut off.

### Steps to Reproduce

1. create a horizontal 100% stacked bar chart (`chart.stacked = true`, `chart.stackType = "100%"`, `plotOptions.bar.horizontal = true`)
2. enabled total-datalabels (`plotOptions.bar.dataLabels.total.enabled = true`)
=> total dataLabel gets cut off

### Expected Behavior

The chart adds a right padding dependent on the max dataLabel width.

### Actual Behavior

The total-dataLabel gets positioned right of the bar and gets cut off, because it overflows.

### Screenshots

![chrome_CTOZi7kLSm](https://user-images.githubusercontent.com/17913239/210218082-41e15e65-3aba-4859-ba46-8bad380f010d.png)

### Reproduction Link

[https://codepen.io/moddesteni/pen/vYaKJQq](https://codepen.io/moddesteni/pen/vYaKJQq)

### Workaround

Defining a manual offset via `grid.padding.right` works as a workaround, but depending on the data it can still get cut off, if a big number is needed to be shown.

Wijlini:
Same here, we need fix for that

nfaycel:
I'm facing the same issue.

junedchhipa:
Fixed and released in **6.10.0** (`da73897d8`).

Verified on 6.10.0 with a horizontal `stackType: '100%'` bar and total labels on, measuring each total label's right edge against the right edge of the SVG:

```
before: +19.6px past the edge   (clipped)
now:    -11.9px                  (inside, with room to spare)
```

The total label sits outside the end of the bar, but on a 100% stack the last bar ends exactly at the plot's right edge, so the label had nowhere to go. The grid now reserves room on the right for it, scoped to exactly that combination (horizontal, stacked, `stackType: '100%'`, totals enabled), so no other chart loses plot width to it.

Guarded by `tests/interaction/specs/stacked-total-datalabels.spec.js`, which asserts the overhang is non-positive rather than asserting a fixed pixel value, so it stays meaningful if the label font or the padding changes.

PR #5315, opened 2026-09-22
ci: run Playwright interaction tests on CI

lovasoa:
Adds a GitHub Actions workflow (`.github/workflows/playwright.yml`) to run the existing Playwright interaction suite (`tests/interaction`, `yarn interaction`) on CI, following the [Playwright CI docs](https://playwright.dev/docs/ci).

- Triggers on `push`/`pull_request` to `main` (matches existing workflows)
- `yarn install --frozen-lockfile` → `yarn build:samples` → `yarn build` (tests load built `dist/` + generated samples over `file://`)
- `npx playwright install --with-deps chromium` (config is chromium-only)
- `yarn interaction --workers=1` (workers=1 per Playwright CI recommendation)
- Uploads `tests/interaction/test-results/` as an artifact

Draft — checking CI results.


junedchhipa:
The workflow does the job, and the run on this PR is the proof: 7 minutes, green, 584 tests that nothing was watching before. Thank you for picking this up.

One request, and it is only about packaging. The `Defaults.js` hunk is a real bug and a good catch, but it is a source fix riding a CI PR. Could it go in its own PR? A CI change is one people want to merge quickly without thinking about chart internals, and that fix deserves a title of its own so it is findable later.

For what it is worth, I measured it and you are right. On `main` the total label clears the SVG viewport by 1.23px; with your change, 11.9px. The reserve gains 10.67px, which is the width of the `%`. So today it is not clipped, it is one font metric away from being clipped.

Two things for when you split it out:

- It needs a test. The existing #3579 guard in `stacked-total-datalabels.spec.js` passes either way, because that case has enough slack to absorb the 10.67px, so nothing currently goes red without your fix.
- The file runs `--workers=4` while the description says `--workers=1`. Either is fine by me. The thing I would add alongside it is `retries: process.env.CI ? 2 : 0` in `tests/interaction/playwright.config.mjs`, which sits at `retries: 0` today. The streaming specs measure frame velocity and are the ones that drift under load, so a single flake would turn the build red. I ran the full suite twice at 4 workers here with no flake, so treat this as a guard rail rather than a problem.

Small one: the other workflows here declare `permissions:`, so `contents: read` would match the house style.

lovasoa:
@junedchhipa I split the PRs, see #5317 :) 

P.S. could you please add a tiny "be concise" to your llm prompt ? 😄

junedchhipa:
Haha. Added 😅 

Reviewing #5317 now, and thanks for splitting it, and for adding the test.

junedchhipa (approved):

